PR #4, opened 2019-11-24
Remove workspace setup

webmaster128:
to make examples self-contained


webmaster128:
Thanks for feedback. Updated. I kept `build_test_all.sh` for development purposes but we could also remove it.

Could you enable the CI for PR builds such that we can see the results of the config change?

ethanfrey:
After some searching, I'm not sure where to enable CI builds for forks, just for branches on this repo. It looks good and I will merge. If you send me a link, I will enable.

PR #10, opened 2019-11-29
Add ERC20 token contract

webmaster128:
TODO:

- [x] Implement `transferFrom`
- [x] Setup CI
- [x] Integration tests
- [x] Convert balance store from JSON to raw uint64
- [x] messages schema generation
- [x] Query handler?


erc20/src/contract/mod.rs
```diff
@@ -0,0 +1,312 @@
...
+            msg: format!(
+                "Insufficient allowance: allowance={}, required={}",
+                allowance, amount
+            ),
+        }
+        .fail();
+    }
+    allowance -= amount;
+    write_allowance(store, &owner_address_raw, &spender_address_raw, allowance);
+    perform_transfer(store, &owner_address_raw, &recipient_address_raw, amount)?;
```
webmaster128: @ethanfrey one of the unittests fails because the allowance update to store is not rolled back when `perform_transfer` fails. Ideas how to test rollbacks in unittests?

webmaster128: Maybe the trick is not to test state at all after failed message executions and just assume nothing changed in this case.

ethanfrey: This is a very good question.
In the higher-level usage, inside the sdk app, we will pass in a KVStore that can be rolled back on error. However, I did not implement such in rust. This would be a very nice feature to add.

It is not just about testing state, but if you test a sequence of commands, some of which succeed and others fail, we want to have the proper result. I created issue https://github.com/confio/cosmwasm/issues/70 with my first ideas on how to add this functionality. Please add some feedback and other ideas there (especially for the API).

ethanfrey:
As to schema generation and query, they are available in cosmwasm 0.5 and maybe the easiest is to make a new project based on https://github.com/confio/cosmwasm-template and then copy your `contract.rs` there (I assume all your real changes were there?), merging in the auto-generated `QueryMsg` and `query` function, as well as the newer `derive` statments on all structs.

If that is too much, you can merge in the updates to `cosmwasm-examples/escrow` piece by piece (which were porting into `cosmwasm-templates`, just simpler business logic there.

erc20/src/contract/tests.rs
```diff
@@ -0,0 +1,795 @@
...
+    params
+}
+
+fn get_name<T: Storage>(store: &T) -> String {
+    let data = store.get(KEY_NAME).expect("no name data stored");
+    let value = String::from_utf8(data).unwrap();
+    return value;
+}
+
+fn get_symbol<T: Storage>(store: &T) -> String {
```
ethanfrey: If you add support for the `query` interface, it would be interesting to have all these implementations make use of it. Either as a `RawQuery`, or custom `QueryMsg` variants for each thing queries. This then models what would be shown to an external client querying. Both for building apps, as well as easily porting logic to integration tests.

The fact you already extracted all this into `get_*` functions make refactoring it much easier.

erc20/src/contract/mod.rs
```diff
@@ -0,0 +1,284 @@
...
+
+use cosmwasm::errors::{ContractErr, DynContractErr, ParseErr, Result};
+use cosmwasm::serde::from_slice;
+use cosmwasm::storage::Storage;
+use cosmwasm::types::{Params, Response};
+
+#[derive(Serialize, Deserialize, Clone, PartialEq)]
+pub struct InitialBalance {
+    pub address: String,
+    pub amount: u64,
```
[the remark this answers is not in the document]
webmaster128: Isn't a u64 <-> JSON number conversion a bug in the serde component?

Storage: Right you are, u64 is not sufficient for Ethereum-style 18 digit decimals.

erc20/src/contract/mod.rs
```diff
@@ -0,0 +1,284 @@
...
+            msg: "Decimals must not exceed 18",
+        }
+        .fail();
+    }
+    store.set(KEY_DECIMALS, &msg.decimals.to_be_bytes());
+
+    // Initial balances
+    let mut total: u64 = 0;
+    for row in msg.initial_balances {
+        let raw_address = parse_20bytes_from_hex(&row.address)?;
```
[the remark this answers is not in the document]
webmaster128: Hmm, this is an issue. We expect smart contracts to process the various printable representations of addresses. As a smart contract developer, I'd expect a normalized address type nativ to the current chain, like [address in Solidity](https://solidity.readthedocs.io/en/v0.5.3/types.html#address) or the 20 byte address in weave. So I'd expect the flexibility to happen on a different level in the stack.

erc20/src/contract/mod.rs
```diff
@@ -0,0 +1,284 @@
...
+            Err(_) => ContractErr {
+                msg: "Corrupted data found. 8 byte expected.",
+            }
+            .fail(),
+        },
+        None => Ok(0u64),
+    };
+}
+
+fn read_allowance<T: Storage>(store: &T, owner: &[u8; 20], spender: &[u8; 20]) -> Result<u64> {
```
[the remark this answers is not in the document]
webmaster128: Prefixing makes sense for sure. This is just the naive approach and sufficient for small contracts where you have key lengths 20, 40 and other (!= 20 and != 40).

erc20/src/contract/mod.rs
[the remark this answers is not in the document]
ethanfrey: I don't know what serde does if it is say 2^62. I assume it writes it all to json.

JSON actually allows larger numbers, just defines a number as `integer fractional exponent` without any rule on how big they may be https://www.crockford.com/mckeeman.html

JS implementation looses precision over 2^53
Go will actually export and parse `int64` up to the maximum it can store.

The json issue here is only really with js interop. And if we are using json to communicate between go and rust, it doesn't fully apply. That said, some json is exposed client side, and it would be good if it was supported by all client libs (esp. js)

erc20/src/contract/mod.rs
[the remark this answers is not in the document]
ethanfrey: I am not sure I agree. 
I would like a contract to be able to run on different chains - one with eg. 20 byte addresses and one with 32 byte addresses.

If we just treat addresses as opaque strings, what is the loss? I would never expect a contract to generate an address, just store them and use addresses it has previously seen.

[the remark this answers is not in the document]
webmaster128: I am semi-active in multiple ecosystems that did or are about to change their user readable address representation. I would just not assume that an address is encoded in the same way all the time. Printable addresses are made for humans, not for machines and they can live at a higher level. I'd expect the contract to run unchanged if the UI layer decides to swicht between hex<->checksummed hex<->bech32<->SS58, as long as the encoded address data remains the same.

However, I am fine with not hardcoding the address length. If we have a constant `ADDRESS_LENGTH` receive parsed bytes as addresses, that would be cool.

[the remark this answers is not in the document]
ethanfrey: Let us have a call on this.
I have quite a few mostly formed thoughts on identifiers, etc. I do understand having one canonical form, but wonder if that is also raw bytes. Maybe it is.

This topic is bigger than the pr. For now, the only place it is embedded uses bech32 encoding, so hex decoding won't work. I'd merge a first version that leaves strings untouched. And then we can discuss the proper solution,  which will likely involve changes to the wasmd runtime as well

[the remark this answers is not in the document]
ethanfrey: Your point about ecosystems changing the human readable format is well taken

erc20/src/contract/mod.rs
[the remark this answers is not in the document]
webmaster128: Aha! Very interesting. I always thought JSON is a subset of JavaScript in syntax as well as semantics. But JSON Number != JS number. Good to know.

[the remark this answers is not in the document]
webmaster128: Should be done in https://github.com/confio/cosmwasm-examples/pull/10/commits/f7b0f7d7164fa15ad31203fa6704b31ad98bcf7d. Please unresolve this comment in case I missed something.

erc20/src/contract/mod.rs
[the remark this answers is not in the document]
webmaster128: I solved this by hashing the address for now: https://github.com/confio/cosmwasm-examples/pull/10/commits/39b689f056ca13701f14130cea12bc1b8b6d57fc The reason for doing so is the desire to have fixed length adresses.

erc20/src/contract/mod.rs
```diff
@@ -0,0 +1,284 @@
...
+fn try_transfer<T: Storage>(
+    store: &mut T,
+    params: Params,
+    recipient: &str,
+    amount: u64,
+) -> Result<Response> {
+    let sender_address_raw = parse_20bytes_from_hex(&params.message.signer)?;
+    let recipient_address_raw = parse_20bytes_from_hex(&recipient)?;
+
+    perform_transfer(store, &sender_address_raw, &recipient_address_raw, amount)?;
```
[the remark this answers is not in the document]
webmaster128: `perform_transfer` is used twice and I'd prefer to keep it separate taking fixed length addresses. Resolving this for now. Feel free to unresolve.

erc20/src/contract/tests.rs
webmaster128: Balance/allowance queries are currently blocked by https://github.com/confio/cosmwasm/issues/71. Even with custom query messages, the return type `QueryResponse` expects UTF-8 compatible keys.

webmaster128: Here we have an initial query: https://github.com/confio/cosmwasm-examples/pull/10/commits/51bdb51db2d7e329c4fde6c76a4883da2507467e I'm not convinced by the overhead this creates.

ethanfrey:
Thank you for the update.
I would open two issues to discuss some of the larger design issues: how queries work, and how addresses are presented.

Otherwise, I think this PR is quite good, but it raises issues we should clarify. Probably the address issue before merge, but if you want to merge, I don't want to block on that either.

erc20/src/contract/tests.rs
ethanfrey: By overhead, you mean the fact you have to execute wasm code to get the value, rather than the go runtime just returning the value from the kvstore?

My idea with query was really to allow arbitrary logic to run (so like top 5 account balances as a query). And RawQuery just as a simple demo. I also start thinking we should support RawQuery inside the runtime and then only send custom queries to the wasm contract.

Query is really a WIP and we should work out a design doc. But I find having a stub makes it easier to iterate on the design. I started this issue to continue a deeper discussion: https://github.com/confio/cosmwasm/issues/72

ethanfrey:
I also added https://github.com/confio/cosmwasm/issues/73 to discuss the canonical address idea. Pleae add

erc20/src/contract/mod.rs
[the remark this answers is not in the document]
webmaster128: I now implemented 3 prefixes: "config", "blances", "allowances" using the length prefixed prefix format.

erc20/src/contract/tests.rs
webmaster128: All queries in this contract here are now raw queries. Additional smart queries can be added that take human readable addresses once https://github.com/confio/cosmwasm/issues/72 and https://github.com/confio/cosmwasm/issues/73 are done

erc20/src/contract/mod.rs
```diff
@@ -0,0 +1,284 @@
...
+ * - ascii("symbol") stores the ticker symbol ([A-Z]{3,6} as ASCII)
+ * - ascii("decimals") stores the fractional digits (unsigned int8)
+ * - ascii("total_supply") stores the total supply (big endian encoded unsigned int64)
+ * - `address` stores balance data (as JSON) for a single address. `address`
+ *   is always 20 bytes long and thus can not conflict with other keys.
+ * - `owner` + `spender` stores allowance data (big endian encoded unsigned int64)
+ *   for an owner spender pair address. `owner` + `spender` is always 40 bytes long
+ *   and thus can not conflict with other keys.
+ */
+pub const KEY_TOTAL_SUPPLY: &[u8] = b"total_supply";
```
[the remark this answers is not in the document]
webmaster128: Since all my storage value are raw data, I want to avoid adding JSON into the mix. A smart query should be able to merge all configs into one result dynamically. Closing for now.

erc20/src/contract/mod.rs
```diff
@@ -71,14 +64,28 @@ pub const KEY_DECIMALS: &[u8] = b"decimals";
...
+
     // Name
     if !is_valid_name(&msg.name) {
         return ContractErr {
             msg: "Name is not in the expected format (3-30 UTF-8 bytes)",
         }
         .fail();
     }
-    store.set(KEY_NAME, msg.name.as_bytes());
+    config_store.set(KEY_NAME, msg.name.as_bytes());
```
ethanfrey: One comment here (which really needs to make it to the docs):

TL;DR - I need to document gas pricing, but storage access is the largest sink, and best to combine multiple keys that are read/written together into one struct, that requires one read or write. In this case, definitely combine `name` and `symbol`. `decimals` likely belongs to the same struct. `total_supply` may belong there as well, unless it is often updated on it's own without the other items, when it would make sense to take a small hit on init to save some gas in future handle operations.

All storage access is calls to an underlying iavl tree in cosmos-sdk (for now.. maybe there are other implementations later, but we should design for this case now). If we [look at the pricing](https://github.com/cosmos/cosmos-sdk/blob/4ffabb65a5c07dbb7010da397535d10927d298c1/store/types/gas.go#L153-L163), we see: 

```go
// KVGasConfig returns a default gas config for KVStores.
func KVGasConfig() GasConfig {
	return GasConfig{
		HasCost:          1000,
		DeleteCost:       1000,
		ReadCostFlat:     1000,
		ReadCostPerByte:  3,
		WriteCostFlat:    2000,
		WriteCostPerByte: 30,
		IterNextCostFlat: 30,
	}
}
```

(note that we divide wasm gas costs by 100 to normalize them to sdk gas costs, and cpu usage for the contracts I tested is usually between 50-100k (or 500-1000 sdk gas), far smaller than even one db read). To optimize contracts it makes sense to minimize storage access before minimizing parse/serialization logic.

3 writes of 10 bytes each = 3*2000 + 30*30 = 6900 gas
1 write of 40 bytes (adding codec overhead) = 2000 + 30*40 = 3200gas

for comparison, parsing the Params struct in wasm costs around 50-100 sdk gas, and doing one round of sha256 costs around 100-150 sdk gas.

erc20/src/contract/mod.rs
```diff
@@ -127,16 +127,9 @@ pub fn handle<T: Storage>(store: &mut T, params: Params, msg: Vec<u8>) -> Result
...
     let msg: QueryMsg = from_slice(&msg).context(ParseErr { kind: "QueryMsg" })?;
-    match msg {
-        QueryMsg::TotalSupply => perform_raw_query(
-            store,
-            RawQuery {
-                key: from_utf8(KEY_TOTAL_SUPPLY).context(Utf8Err {})?.to_string(),
-            },
-        ),
-    }
+    match msg {}
```
ethanfrey: Does this compile? It doesn't seem to return anything...

I would explicitly replace the function body `ContractErr{ msg: "query unimplemented" }.fail()`

TODO: ethan - more docs on error handling

erc20/src/contract/mod.rs
```diff
@@ -227,7 +220,8 @@ fn perform_transfer<T: Storage>(
     to: &[u8; 32],
     amount: u128,
 ) -> Result<()> {
-    let mut from_balance = read_balance(store, from)?;
+    let mut balances_store = PrefixedStorage::new(store, PREFIX_BALANCES);
```
ethanfrey: Very nice, this looks clean

erc20/src/contract/mod.rs
```diff
@@ -281,13 +275,10 @@ pub fn parse_u128(decimal: &str) -> Result<u128> {
     }
 }
 
-fn read_balance<T: Storage>(store: &T, owner: &[u8; 32]) -> Result<u128> {
-    return read_u128(store, owner);
-}
-
-fn read_allowance<T: Storage>(store: &T, owner: &[u8; 32], spender: &[u8; 32]) -> Result<u128> {
+fn read_allowance<T: Storage>(store: &mut T, owner: &[u8; 32], spender: &[u8; 32]) -> Result<u128> {
+    let allowances_store = PrefixedStorage::new(store, PREFIX_ALLOWANCES);
```
ethanfrey: Thinking of your NamespacesStorage branch, and seeing this usecase, what you are doing in the concat is basically nested `PrefixStorage`. What about using nested `PrefixedStorage` to represent this, like:

```rust
let allowances_store = PrefixedStorage::new(store, PREFIX_ALLOWANCES);
let owners_allowances = PrefixedStorage::new(&mut allowances_store, owner);
read_u128(&owners_allowances, &spender)
```
(in rust, it is typical is to omit `return` and the trailing `;` when it is the last statement in a block)

Also note that nesting PrefixedStorage with any of your non-collision schemes, means that we don't need to assert the length of the addresses in the first place. They can be any length, they must just be unique (no two users with same address).

ethanfrey: Random thoughts on API... Maybe a final PrefixedStorage could make multi-level prefixes in one step. Rust does not support [variadic functions, just variadic macros](https://doc.rust-lang.org/stable/rust-by-example/macros/variadics.html). So the straight-forward approach would be to place the multiple keys in one slice (which is what variadics often do behind the scenes in other languages):

```rust
let allowances = PrefixedStorage::multilevel(store, &[PREFIX_ALLOWANCES, owner]);
read_u128(&allowances, &spender)
```

One could also make some macro for this if it is too verbose, but would prefer to avoid macros unless they bring a lot of benefit. For me this just obfuscates the code flow a bit:

```rust
let allowances = prefix!(store, PREFIX_ALLOWANCES, owner);
read_u128(&allowances, &spender)
```

What do you think?

erc20/src/contract/tests.rs
```diff
@@ -12,41 +12,70 @@ fn mock_params_height(signer: &str, height: i64, time: i64) -> Params {
 }
 
 fn get_name<T: Storage>(store: &T) -> String {
-    let data = store.get(KEY_NAME).expect("no name data stored");
-    let value = String::from_utf8(data).unwrap();
-    return value;
+    let key = [
```
ethanfrey: Is there a reason you don't use `PrefixedStorage` here? But rather manually implement the key prefixing? Since this is test code, efficiency doesn't seem a reason.

I think it would be easier to read like:

```rust
fn get_name<T: Storage>(store: &T) -> String {
  let prefixed = PrefixedStorage::new(store, KEY_NAME);
   let data = prefixed.get(&key).expect("no name data stored");
   return String::from_utf8(data).unwrap();
}
```

I'm sure you know this, but the panics (expect, unwrap) are great for test code to get stack traces, but should be avoided at all costs in the production code (what is compiled to wasm), as they just abort the contract and give no usable error message to help debugging.

erc20/src/contract/mod.rs
webmaster128: Yes, this compiles. But the gate keeper should be the line above

```rust
let msg: QueryMsg = from_slice(&msg).context(ParseErr { kind: "QueryMsg" })?;
```

for

```rust
pub enum QueryMsg {}
```

I can't think of any JSON value that would successfully deserialize.

erc20/src/contract/tests.rs
webmaster128: The reson was (which is one of the most annoying things in the current `PrefixedStorage` implementation): in the snippet you posted, `store` needs to be mutable, even if we never ready anything from the store. This is because a mutable reference is stored in PrefixedStorage just in case we need to `.set`. Any idea how to improve the implementation to avoid that?

erc20/src/contract/mod.rs
webmaster128: I was thinking about this too at some point but am not confident that we can avoid collisions by nesting `PrefixedStorage`s. I need to think about this very carefully again. It would create a very good API though.

erc20/src/contract/mod.rs
webmaster128: Thanks for the insights. I was not expecting DB reads to be so expensive. Writes don't matter much in this case as this happens only once in init. Read cost surprises me, so 👍 for having that nicely documented.

erc20/src/contract/mod.rs
ethanfrey: Serde json ignores extra fields.
It requires all present fields unless they are wrapped in Option<> to be consistent with rust type system.

webmaster128: Yes, but `QueryMsg` is an enum with no cases, not an empty struct. There is no way an instance `msg: QueryMsg` can ever exist.

erc20/src/contract/mod.rs
ethanfrey: I was thinking of your length prefixed implementation.

And if the first prefix storage only had prefix storages as direct children? This is a requirement in any case.

Prefix foo, key bar and
Prefix fo key obar
Will never collide.

But a hand generated \03foobar will collide.

Nested prefixed storage should work with the same caveat

erc20/src/contract/mod.rs
ethanfrey: Touche. You are right

erc20/src/contract/tests.rs
ethanfrey: Good point. Off the top of my head, I don't see how to do so. Maybe I can come up with a way. I think it would require to separate out Storage and ReadOnlyStorage as different traits and allow a struct to just implement the ReadOnly version.

I don't think it is possible to return an object that cannot be mutated. 

I will think more on this

webmaster128: Given the strictness by default approach from Rust, we might want to consider the following interface (that makes read-only the default by naming)

```
// A read-only storage
pub trait Storage {
    fn get(&self, key: &[u8]) -> Option<Vec<u8>>;
}

// A read-write storage
pub trait RwStorage: Storage {
    fn set(&mut self, key: &[u8], value: &[u8]);
}
```

webmaster128: I tried

```rust
pub trait ReadonlyStorage {
    fn get(&self, key: &[u8]) -> Option<Vec<u8>>;
}

pub trait Storage: ReadonlyStorage {
    fn set(&mut self, key: &[u8], value: &[u8]);
}
```

but this cannot be added in a compatible manner. You cannot just implement `Storage` as before due to the lack of inference. The syntax above creates a [supertrait](https://doc.rust-lang.org/rust-by-example/traits/supertraits.html#supertraits) that needs to be implemented as follows:

```rust
impl ReadonlyStorage for MockStorage {
    fn get(&self, key: &[u8]) -> Option<Vec<u8>> {
        self.data.get(key).cloned()
    }
}

impl Storage for MockStorage {
    fn set(&mut self, key: &[u8], value: &[u8]) {
        self.data.insert(key.to_vec(), value.to_vec());
    }
}
```

ethanfrey:
Nice addition of schema.

I think there still a discussion needed about how the query handler works, and this will be overhauled for 0.6, so I would leave it without a query handler for now, and we add a proper one later.

erc20/src/contract/tests.rs
webmaster128: Done in 7bc32fc022bdd678983eaaaef95739ff2b054c7a

erc20/src/contract/mod.rs
webmaster128: This is suggestion implemented in f4a68175968ceb1994a0a78f0084c7fd396a0b79. Nested `PrefixedStorage` should work as long as you stick with one key composition type at every namespace.

- Root namespace `:`: every key must be a TypeX (e.g. wrapper like PrefixedStorage or plain key)
- Sub namespace 1 `:foo:`: everything in `:foo:` must be a TypeY (e.g. wrapper like PrefixedStorage or plain key)
- Sub namespace 1 `:bar:`: everything in `:bar:` must be a TypeX (e.g. wrapper like PrefixedStorage or plain key)

`:foo:prefixed(moep)`
`:foo:prefixed(hello)`
`:foo:prefixed(\x03sun)`
`:bar:plain(\x04moep)`
`:bar:plain(hello)`
`:bar:plain(sun)`

are all good.

Collisions can only occur of you don't stick to the same wrapper type at one namespace. E.g. use `PrefixedStorage` and plain key inside of namespace `:foobar:`.

`:foobar:prefixed(moep)`
`:foobar:plain(\x04moep)`

erc20/src/contract/mod.rs
webmaster128: Done in https://github.com/confio/cosmwasm-examples/pull/10/commits/a3eaaebfb19de02ce21566248ed739437d5e1e8b. I merged all "constants" and left total suppy separate. This should show the idea.

erc20/Cargo.toml
```diff
@@ -0,0 +1,41 @@
...
+[features]
+default = ["cranelift"]
+# for quicker tests, cargo test --lib
+# for more explicit tests, cargo test --features=backtraces
+backtraces = [ "cosmwasm/backtraces", "cosmwasm-vm/backtraces" ]
+cranelift = [ "cosmwasm-vm/default-cranelift"]
+singlepass = [ "cosmwasm-vm/default-singlepass"]
+
+[dependencies]
+cosmwasm = { git = "https://github.com/confio/cosmwasm", rev = "b0f9865ccaa4a6a1d943773999b4655a3df73843" }
```
ethanfrey: Let's update this to 0.6.0 once I release that, and before merging this pr

erc20/Cargo.toml
```diff
@@ -0,0 +1,41 @@
...
+schemars = "0.5"
+serde = { version = "1.0.60", default-features = false, features = ["derive"] }
+snafu = { version = "0.5.0", default-features = false, features = ["rust_1_30"] }
+hex = "0.4.0"
+# needed for wasm-pack build process
+wasm-bindgen = "0.2"
```
ethanfrey: Maybe fix this to 0.2.55 (so we can use most recent cosmwasm-opt:v0.5.2)

erc20/README.md
```diff
@@ -0,0 +1,5 @@
+# An ERC20 token contract
+
+This is an implementation of Ethereum's [ERC20](https://eips.ethereum.org/EIPS/eip-20) interface.
```
ethanfrey: Nice intro.

Can you also copy a few of the sections from `cosmwasm-template`, at least for a quick start (`cargo wasm`, `cargo test`, ...) and a link to other sites for more info?

Or should we include all the generic startup stuff in the top-level of `cosmwasm-examples` and just have this explain the contract and link up to the top-level README.md explaining how to run them?

erc20/examples/schema.rs
```diff
@@ -0,0 +1,42 @@
...
+    let schema = schema_for!(InitMsg);
+    export_schema(&schema, &pwd, "init_msg.json");
+
+    let schema = schema_for!(HandleMsg);
+    export_schema(&schema, &pwd, "handle_msg.json");
+
+    let schema = schema_for!(QueryMsg);
+    export_schema(&schema, &pwd, "query_msg.json");
+
+    let schema = schema_for!(BalanceResponse);
```
ethanfrey: This is nice. We really should come up with a standard for how to collect all the various query responses. Ideally associating them somehow with the query request enums...

ethanfrey: (That is out of scope... just a thought on future extensions to cosmwasm, so any thoughts you come on making queries nice... please share)

erc20/src/contract/mod.rs
```diff
@@ -0,0 +1,388 @@
...
+
+use cosmwasm::errors::{ContractErr, DynContractErr, ParseErr, Result, SerializeErr};
+use cosmwasm::serde::{from_slice, to_vec};
+use cosmwasm::traits::{Api, Extern, ReadonlyStorage, Storage};
+use cosmwasm::types::{CanonicalAddr, HumanAddr, Params, Response};
+
+#[derive(Serialize, Deserialize, Clone, PartialEq, JsonSchema)]
+pub struct InitialBalance {
+    pub address: HumanAddr,
+    pub amount: String,
```
ethanfrey: Nice use as String.
Can you add a comment what that means - is it an integer? May it include decimals?

If `decimals` is 5 and `amount` is "12345", do I have 12345 units or 0.12345?

erc20/src/contract/mod.rs
```diff
@@ -0,0 +1,388 @@
...
+        recipient: HumanAddr,
+        amount: String,
+    },
+}
+
+#[derive(Serialize, Deserialize, Clone, Debug, PartialEq, JsonSchema)]
+#[serde(rename_all = "lowercase")]
+pub enum QueryMsg {
+    Balance {
+        address: HumanAddr,
```
ethanfrey: :smile:

erc20/src/contract/mod.rs
```diff
@@ -0,0 +1,388 @@
...
+) -> Result<Response> {
+    let msg: InitMsg = from_slice(&msg).context(ParseErr { kind: "InitMsg" })?;
+
+    let mut total_supply: u128 = 0;
+    {
+        // Initial balances
+        let mut balances_store = PrefixedStorage::new(&mut deps.storage, PREFIX_BALANCES);
+        for row in msg.initial_balances {
+            let raw_address = deps.api.canonical_address(&row.address)?;
+            let amount_raw = parse_u128(&row.amount)?;
```
ethanfrey: Okay, the amounts are all unsigned integers. Fair enough. This should just be in a comment above

erc20/src/contract/mod.rs
```diff
@@ -0,0 +1,388 @@
...
+
+    let mut config_store = PrefixedStorage::new(&mut deps.storage, PREFIX_CONFIG);
+    let constants = to_vec(&Constants {
+        name: msg.name,
+        symbol: msg.symbol,
+        decimals: msg.decimals,
+    })
+    .context(SerializeErr { kind: "Constants" })?;
+    config_store.set(KEY_CONSTANTS, &constants);
+    config_store.set(KEY_TOTAL_SUPPLY, &total_supply.to_be_bytes());
```
ethanfrey: very nice splitting of concerns

erc20/src/contract/mod.rs
```diff
@@ -0,0 +1,388 @@
...
+    }
+}
+
+pub fn query<S: Storage, A: Api>(deps: &Extern<S, A>, msg: Vec<u8>) -> Result<Vec<u8>> {
+    let msg: QueryMsg = from_slice(&msg).context(ParseErr { kind: "QueryMsg" })?;
+    match msg {
+        QueryMsg::Balance { address } => {
+            let address_key = deps.api.canonical_address(&address)?;
+            let balance = read_balance(&deps.storage, &address_key)?;
+            let out = to_vec(&BalanceResponse {
```
ethanfrey: This is very proper. Reading through it, I think it could be a bit shorter (but that means changing cosmwasm).

What if there was a function `to_json`, such that:

`let out = to_json("BalanceResponse", &BalanceResponse{ balance: balance.to_string() })?;`

was valid. Basically collapse in the SerializeErr? Or maybe there is someway to auto-derive the type name from serde

If you like the idea, please make an issue for 0.7, otherwise, ignore

erc20/src/contract/mod.rs
```diff
@@ -0,0 +1,388 @@
...
+    recipient: &HumanAddr,
+    amount: &str,
+) -> Result<Response> {
+    let spender_address_raw = &params.message.signer;
+    let owner_address_raw = deps.api.canonical_address(owner)?;
+    let recipient_address_raw = deps.api.canonical_address(recipient)?;
+    let amount_raw = parse_u128(amount)?;
+
+    let mut allowance = read_allowance(&deps.storage, &owner_address_raw, &spender_address_raw)?;
+    if allowance < amount_raw {
```
ethanfrey: Rust compiles code so it should panic on under/overflow. Since these are u128, this would trigger that case.

For for testing, I'd be curious what would happen if you removed this if statement. You should leave it in for explicitness, but good if the framework provides protection against this

erc20/src/contract/mod.rs
```diff
@@ -0,0 +1,388 @@
...
+                "Insufficient funds: balance={}, required={}",
+                from_balance, amount
+            ),
+        }
+        .fail();
+    }
+
+    let mut to_balance = read_u128(&balances_store, to.as_bytes())?;
+
+    from_balance -= amount;
```
ethanfrey: I am pretty sure there is a subtle bug here.
Try setting to and from to the same addresss...

If I send 50 to myself, I believe my balance will increase by 50.

Changing order or reads and writes will fix this, and a clear comment is good.

from: read, check, deduct, store
to: read, add, store

erc20/tests/integration.rs
```diff
@@ -0,0 +1,241 @@
...
+    let owner = address(0);
+    let spender = address(1);
+
+    // Before
+    deps.with_storage(|storage| {
+        assert_eq!(get_allowance(&deps.api, storage, &owner, &spender), 0);
+    });
+
+    // Approve
+    let approve_msg = to_vec(&HandleMsg::Approve {
```
ethanfrey: What do you think of https://github.com/confio/cosmwasm/pull/97 to simplify this syntax?

erc20/Cargo.toml
webmaster128: Done in a63fe5659ae05911a6d681a857d268f32c3b4e7a

erc20/tests/integration.rs
webmaster128: Nice, done in 0756a8ee34f471da5b4830c1f71ccfa501d24335

erc20/src/contract/mod.rs
webmaster128: Yeah, I think the reason to have this check was that I did not know I can test for _panic with message_ at that time

erc20/src/contract/mod.rs
webmaster128: Oh wow, nice catch!

webmaster128: Should be fixed in f18db36ebb7f69369e1b67f2623b2ec4bfc73b76

erc20/src/contract/mod.rs
webmaster128: Well, there is no such thing as positional arguments in JSON. `Balance (HumanAddr)` was nice in code but not in the JSON message

erc20/Cargo.toml
webmaster128: updated to 0.6.0

erc20/examples/schema.rs
webmaster128: With something like https://github.com/carllerche/codegen, it would be possible to define the JSON structure in some kind of data file and generate both rust code and schemas in one step from a single source.

ethanfrey:
Looks good and ready to merge. Anything blocking a merge in your opinion?

After it is on master, we can publish this to cargo

PR #13, opened 2019-12-20
Add CI step "Ensure checked-in schemas are up-to-date"

webmaster128:


webmaster128:
seems to work as expected

PR #14, opened 2019-12-21
Check  cargo fmt in CI

webmaster128:


ethanfrey:
Sure. Maybe this belongs in `cosmwasm` for real.

And you can add `cargo clippy` to do a much deeper linting.
And `cargo clippy -- -W clippy::pedantic` to be absolutely annoying.

PR #18, opened 2019-12-30
Update escrow to cosmwasm 0.6

### LINKED ISSUE #8: Fix README file in escrow
This is the default README from cosmwasm-template.

Let's make proper info tied to escrow

### LINKED ISSUE #17: Update escrow to 0.6
* Update the contract to work well with the 0.6 cosmwasm release.
* Update tutorial to check out v0.5.2 tag

ethanfrey:
Closes #17 
Closes #8 

Added [comment to the tutorial](https://github.com/cosmwasm/docs/blob/master/docs/getting-started/editing-escrow-contract.md#editing-code) about using the `escrow-0.1.0` tag of the examples. 

Once we integrate 0.6 into wasmd, we need to properly update the tutorial.

Question: should I add a custom query function? What makes sense? `is_expired` was my thought, but we cannot do that with current query api (it has no params).


webmaster128:
> What makes sense? is_expired was my thought, but we cannot do that with current query api (it has no params).

This sounds like a great use case we don't have yet: query enum with no argument. I don't see why this should not work. We need to be able to support such queries one way or another. How does

```rust
#[derive(Serialize, Deserialize, Clone, Debug, PartialEq, JsonSchema)]
#[serde(rename_all = "lowercase")]
pub enum QueryMsg {
    IsExpired,
}
```

serialize? Like this?

```json
{ "is_expired": {} }
```

escrow/LICENSE
```diff
@@ -189,7 +189,7 @@ License: Apache2.0
       same "printed page" as the copyright notice for easier
       identification within third-party archives.
 
-   Copyright 2019 Confio UO
+   Copyright 2019 Ethan Frey <ethanfrey@users.noreply.github.com>
```
webmaster128: I think we should avoid manipulating the Apche 2 LICENSE. This section is not meant to be modified. It is a template for a NOTICE file. See e.g. https://github.com/iov-one/iov-core/blob/master/NOTICE (in this repo we also modified LICENSE which is not necessary).

escrow/rustfmt.toml
```diff
@@ -0,0 +1,15 @@
...
+# unstable... should we require `rustup run nightly cargo fmt` ?
+# or just update the style guide when they are stable?
+#fn_single_line = true
+#format_code_in_doc_comments = true
+#overflow_delimited_expr = true
+#reorder_impl_items = true
+#struct_field_align_threshold = 20
+#struct_lit_single_line = true
+#report_todo = "Always"
+
```
webmaster128: Why do we need this file? Don't we use all defaults everywhere?

escrow/LICENSE
webmaster128: Implemented for the template in https://github.com/confio/cosmwasm-template/pull/19

webmaster128:
If I did not miss anything here, it is possible to approve or cancel this escrow multiple times, since the state is not stored. In contract to the weave implementation, we can't delete the escrows after one of those operations happened, right?

escrow/rustfmt.toml
ethanfrey: There are some things I did not like about the defaults.
In particular, `.context(SerializeError{ kind: "State"})` should not need 3 lines.

It turns out I can customize `rustfmt` to keep that on one line :smile: but those flags are only available on `nightly` :frowning_face: 

In the end, I just enforced the unix line ending and tabs as spaces. Same thing as in editorconfig, but this way the CI will fail on a bad line ending.

ethanfrey:
> >    What makes sense? is_expired was my thought, but we cannot do that with current query api (it has no params).

> This sounds like a great use case we don't have yet: query enum with no argument. I don't see why this should not work. We need to be able to support such queries one way or another. How does

```rust
#[derive(Serialize, Deserialize, Clone, Debug, PartialEq, JsonSchema)]
#[serde(rename_all = "lowercase")]
pub enum QueryMsg {
    IsExpired,
}
```

> serialize? Like this?

> { "is_expired": {} }

I think it is `"is_expired"`, which is not even a dictionary. To get that serialization, I would use:

```rust
#[derive(Serialize, Deserialize, Clone, Debug, PartialEq, JsonSchema)]
#[serde(rename_all = "lowercase")]
pub enum QueryMsg {
    IsExpired{},
}
```

(note the `{}` after IsExpired) this keeps it more consistent.

The big issue however, is that currently the query interface has no idea what the current block height or time is, so it cannot compare. We could add that to the query interface, but first in 0.7 (I was wondering what those values should be when calling a smart contract query in the middle of a block... also we have no values like `msg.sender` and I do not want the overhead of providing the contract balance for a simple query, so I would not like to add the entire params. Maybe just a subset?

ethanfrey:
> If I did not miss anything here, it is possible to approve or cancel this escrow multiple times, since the state is not stored. In contract to the weave implementation, we can't delete the escrows after one of those operations happened, right?

The approve multiple times is desired. As we allow partial approvals (this is not an atomic swap, an arbiter may elect to break it into 4 smaller payments, then maybe not release the final one).

Closing (return) multiple times is somewhat pointless, but not dangerous. I could add an `is_closed` flag on the contract and return error if closed a second time. We do allow people to top up an escrow, so I would not close it until it is expired, as there may be future payments planned.

In any case, this is a demo... we really do need to consider a multi-escrow contract if we were going to use this for real

escrow/LICENSE
ethanfrey: I fixed this in https://github.com/confio/cosmwasm-examples/pull/18/commits/6503aa64880a339ffcfe908184f9b2f75d253c80

escrow/Developing.md
```diff
@@ -0,0 +1,143 @@
...
+
+However, we don't just want to test the logic rust, but also the compiled Wasm artifact
+inside a VM. You can look in `tests/integration.rs` to see some examples there. They
+load the Wasm binary into the vm and call the contract externally. Effort has been
+made that the syntax is very similar to the calls in the native rust contract and
+quite easy to code. In fact, usually you can just copy a few unit tests and modify
+a few lines to make an integration test (this should get even easier in a future release).
+
+To run the latest integration tests, you need to explicitely rebuild the Wasm file with
+`cargo wasm` and then run `cargo integration-tests`.
```
webmaster128: This must be `cargo integration-test` (no trailing s)

PR #30, opened 2020-01-23
[WIP] Add nameservice contract

clevinson:
Current draft of nameservice contract. Would appreciate feedback / thoughts.

In general, this seems like a good basic implementation... works fine, basic test coverage, etc.

Happy to make this a bit more involved if you'd like to see more functionality (e.g. separating out "resolved address" from "owner", so they don't have to be the same).

Some questions I have:
- could add validation on acceptable names, same for Config {name}... (not allow empty string... maybe require a minimum length?)
- seems a bit strange that to resolve from canonical_address to human_address, you use functions in deps.api. Is there a specific reason the address structs don't implement the standard `From` or `Into` traits for eachother?
- should be "state" be included as a schema in examples/schema.rs? I saw some examples where this was the case, and some examples where it was not.
- when writing tests, what is the significance of the two seperate coin arguments in the `mock_params` function (send, balance)?


ethanfrey:
> seems a bit strange that to resolve from canonical_address to human_address, you use functions in deps.api. Is there a specific reason the address structs don't implement the standard From or Into traits for eachother?

Good question and this should be documented. Both items are blockchain specific. In particular, cosmwasm is expected to be run in multiple cosmos-sdk chains with different bech32 prefixes soon. We cannot translate without knowing the prefix. 

I originally wanted to add all that logic in a wasmd library, but was convinced that this logic actually belongs in the blockchain itself (what if we embed into a chain that uses ethermint and uses checksum hex addresses?). The contract should be written in a way to easily run in multiple blockchains (a big goal of cosmwasm is interoperability), so this was pushed out into the runtime.

Does that answer your question? Do you have an idea how to communicate that better (in tutorials, docs, etc)?

ethanfrey:
> should be "state" be included as a schema in examples/schema.rs? I saw some examples where this was the case, and some examples where it was not.

All json structs you want to expose should be in `schema.rs`. Whether you consider state part of your public api (raw reads) or only want people to interact with smart queries (and add those queries + response to schema.rs) is a decision of every contract owner. For simple contracts, raw reads are probably good. For complex ones, better to use smart queries (eg. erc20 get allowance by granter and grantee)

ethanfrey:
> when writing tests, what is the significance of the two seperate coin arguments in the mock_params function (send, balance)?

Send is how many tokens are send in this transaction. If you require payment in native token (eg. sell your name), the contact can require a minimum amount for `params.message.sent_funds` or fail with some "insufficient payment" message. `params.contract.balance` shows the current funds on the contracts account. We use this in eg. escrow to be able to release all funds held by the contract. 

This part definitely need some more docs. And maybe helpers to enforce minimum payment or make a method "unpayable" (reject all calls that send tokens).

ethanfrey:
Good questions. I hope my answers above make sense ^^

I would be very happy if you can put those in a sensible place in the docs in a form that makes sense to you (and more likely to other people new to cosmwasm). I will review any PR and fix it up to ensure it is technically correct, so don't worry about being exactly correct, more being understandable

nameservice/Cargo.toml
```diff
@@ -0,0 +1,44 @@
...
+[features]
+default = ["cranelift"]
+# for quicker tests, cargo test --lib
+# for more explicit tests, cargo test --features=backtraces
+backtraces = [ "cosmwasm/backtraces", "cosmwasm-vm/backtraces" ]
+cranelift = [ "cosmwasm-vm/default-cranelift"]
+singlepass = [ "cosmwasm-vm/default-singlepass"]
+
+[dependencies]
+cosmwasm = { version = "~0.6.0" }
```
ethanfrey: We really need to make 0.6.3 the default in template. Can you update to that and use some of the modern error helpers?

ethanfrey: Although I think cw-storage will pull in 0.6.3. Anyway, good to be explicit about that here

nameservice/Importing.md
```diff
@@ -0,0 +1,114 @@
...
+To generate the schema:
+
+```sh
+cargo run --example schema
+```
+
+And to generate the `contract.wasm` and `hash.txt`:
+
+```sh
+docker run --rm -u $(id -u):$(id -g) -v $(pwd):/code confio/cosmwasm-opt:0.4.1
```
ethanfrey: More outdated code. See issue https://github.com/confio/cosmwasm-template/issues/25

Can you please update the text here and make a PR for fixing the template?

nameservice/schema/count_response.json
```diff
@@ -0,0 +1,14 @@
+{
```
ethanfrey: This came from the demo and should probably be deleted somehow.
`cargo examples` cannot do the rm (it's not a makefile).

Do you have an idea how to automate cleanup of these files? In `cosmwasm-opt` build script I guess, but maybe in a local command as well.

nameservice/src/msg.rs
```diff
@@ -0,0 +1,28 @@
+use cosmwasm::types::HumanAddr;
+use schemars::JsonSchema;
+use serde::{Deserialize, Serialize};
+
+#[derive(Serialize, Deserialize, Clone, Debug, PartialEq, JsonSchema)]
+pub struct InitMsg {
+    pub name: String
+}
+
+#[derive(Serialize, Deserialize, Clone, Debug, PartialEq, JsonSchema)]
```
ethanfrey: Nice definitions for messages here.

nameservice/src/msg.rs
```diff
@@ -0,0 +1,28 @@
+use cosmwasm::types::HumanAddr;
+use schemars::JsonSchema;
+use serde::{Deserialize, Serialize};
+
+#[derive(Serialize, Deserialize, Clone, Debug, PartialEq, JsonSchema)]
```
ethanfrey: I will look above, but what is the purpose of a name in Init? I think this can be empty for now. Otherwise, use it to initialize some global config. (About to read `contract.rs`, so maybe I adjust this)

nameservice/src/state.rs
```diff
@@ -0,0 +1,33 @@
...
+pub static CONFIG_KEY: &[u8] = b"config";
+
+#[derive(Serialize, Deserialize, Clone, Debug, PartialEq, JsonSchema, NamedType)]
+pub struct NameRecord {
+    pub owner: CanonicalAddr,
+}
+
+#[derive(Serialize, Deserialize, Clone, Debug, PartialEq, JsonSchema, NamedType)]
+pub struct Config {
+    pub name: String
```
ethanfrey: Can you add a comment what this `name` does? Here and in `InitMsg`.

nameservice/src/state.rs
```diff
@@ -0,0 +1,33 @@
...
+
+pub fn config<S: Storage>(storage: &mut S) -> Singleton<S, Config> {
+    singleton(storage, CONFIG_KEY)
+}
+
+pub fn resolver<S: Storage>(storage: &mut S) -> Bucket<S, NameRecord> {
+    bucket(NAME_RESOLVER_KEY, storage)
+}
+
+pub fn resolver_read<S: Storage>(storage: &S) -> ReadonlyBucket<S, NameRecord> {
```
ethanfrey: You can use `<S: ReadonlyStorage>` here, which is easier to fulfill.
Right now we pass `&Storage` into query, but I do hope to refactor this to `&ReadonlyStorage` in the future

nameservice/src/contract.rs
```diff
@@ -0,0 +1,280 @@
...
+    params: Params,
+    name: String,
+) -> Result<Response> {
+    let key = name.as_bytes();
+
+    let record = NameRecord {
+        owner: params.message.signer,
+    };
+
+    if let None = resolver(&mut deps.storage).may_load(key)? {
```
ethanfrey: Minor nit. I would just call `resolver` once:

```rust
let names = resolver(&mut deps.storage);
if let None = names.may_load(key)? {
  names.save(key, &record);
} else {
  // ....
}
```

nameservice/src/contract.rs
```diff
@@ -0,0 +1,280 @@
...
+pub fn try_transfer<S: Storage, A: Api>(
+    deps: &mut Extern<S, A>,
+    params: Params,
+    name: String,
+    to: HumanAddr,
+) -> Result<Response> {
+    let key = name.as_bytes();
+    let new_owner = deps.api.canonical_address(&to)?;
+
+    resolver(&mut deps.storage).update(key, &|mut record| {
```
ethanfrey: Here you only call resolver once, so it makes sense not to use an inner variable

nameservice/src/contract.rs
```diff
@@ -0,0 +1,280 @@
...
+    let record = NameRecord {
+        owner: params.message.signer,
+    };
+
+    if let None = resolver(&mut deps.storage).may_load(key)? {
+        // name is available
+        resolver(&mut deps.storage).save(key, &record)?;
+    } else {
+        // name is already taken
+        ContractErr {
```
ethanfrey: With cosmwasm 0.6.3, you can do `return contract_err("Name is already taken");` which should be easier to read (and write).

nameservice/src/contract.rs
```diff
@@ -0,0 +1,280 @@
...
+    params: Params,
+    name: String,
+    to: HumanAddr,
+) -> Result<Response> {
+    let key = name.as_bytes();
+    let new_owner = deps.api.canonical_address(&to)?;
+
+    resolver(&mut deps.storage).update(key, &|mut record| {
+        if params.message.signer != record.owner {
+            Unauthorized {}.fail()?;
```
ethanfrey: With cosmwasm 0.6.3, you can do `return unauthorized();` which should be easier to read (and write).

nameservice/src/contract.rs
```diff
@@ -0,0 +1,280 @@
...
+        }
+
+        record.owner = new_owner.clone();
```
ethanfrey: Nit: I think you don't need the `.clone()` as we never use `new_owner` again.
More for simplicity to read than the runtime overhead

nameservice/src/contract.rs
```diff
@@ -0,0 +1,280 @@
...
+
+fn query_resolver<S: Storage, A: Api>(deps: &Extern<S, A>, name: String) -> Result<Vec<u8>> {
+    let key = name.as_bytes();
+
+    let record = resolver_read(&deps.storage).load(key)?;
+    let address = deps.api.human_address(&record.owner)?;
+
+    let resp = ResolveRecordResponse { address };
+
+    to_vec(&resp).context(SerializeErr {
```
ethanfrey: If you derive `NamedType` on `ResolveRecordResponse`, then you can do:

```rust
cw_storage::serialize(&resp)
```

And it handles this boilerplate.

nameservice/src/contract.rs
```diff
@@ -0,0 +1,280 @@
...
+
+#[cfg(test)]
+mod tests {
+    use super::*;
+    use cosmwasm::errors::Error;
+    use cosmwasm::mock::{dependencies, mock_params, MockApi, MockStorage};
+    use cosmwasm::serde::from_slice;
+    use cosmwasm::types::coin;
+
+    fn mock_init_and_alice_registers_name(mut deps: &mut Extern<MockStorage, MockApi>) {
```
ethanfrey: I like this a lot to pull out repetitive setup

nameservice/src/contract.rs
```diff
@@ -0,0 +1,280 @@
...
+        // bob can't register the same name
+        let params = mock_params(&deps.api, "bob_key", &coin("2", "token"), &[]);
+        let msg = HandleMsg::Register {
+            name: "alice".to_string(),
+        };
+        let res = handle(&mut deps, params, msg);
+
+        match res {
+            Ok(_) => panic!("Must return error"),
+            Err(Error::ContractErr { msg, .. }) => assert_eq!(msg, "Name is already taken"),
```
ethanfrey: :+1: to ensure they why of the failure

nameservice/src/contract.rs
```diff
@@ -0,0 +1,280 @@
...
+            Err(_) => panic!("Unknown error"),
+        }
+        // alice can't register the same name again
```
ethanfrey: Good check

nameservice/src/contract.rs
```diff
@@ -0,0 +1,280 @@
...
+        let params = mock_params(&deps.api, "alice_key", &coin("2", "token"), &[]);
+        let msg = HandleMsg::Transfer {
+            name: "alice".to_string(),
+            to: HumanAddr::from("bob_key"),
+        };
+
+        let _res =
+            handle(&mut deps, params, msg).expect("contract successfully handles Transfer message");
+        // querying for name resolves to correct address (bob_key)
+        let res = query(
```
ethanfrey: These lines... make `QueryMsg::ResolveRecord`, query, parse `ResolveRecordResponse`, check equal are done many times. Maybe a good helper to make. You could add `assert_name_owner` and then just turn this into one line. Same line in other tests:

`assert_name_owner(&mut deps, "alice", "bob_key");`

nameservice/src/contract.rs
```diff
@@ -0,0 +1,280 @@
...
+            &deps,
+            QueryMsg::ResolveRecord {
+                name: "alice".to_string(),
+            },
+        );
+
+        match res {
+            Ok(_) => panic!("Must return error"),
+            Err(Error::NotFound { kind, .. }) => assert_eq!(kind, "NameRecord"),
```
ethanfrey: Nice check

nameservice/schema/count_response.json
clevinson: will think on this

nameservice/src/msg.rs
clevinson: Didn't have an explicit purpose, I just thought it would be nice to provide some way to differentiate between multiple nameservice contracts by allowing them to have a name.

I'll drop this for now, if we want to add it back eventually, we can do so alongside a more complete `InitMsg` when adding the fee structure in a future PR.

PR #31, opened 2020-01-24
[WIP]: Add purchase and transfer fees to nameservice contract

clevinson:
First attempt at adding purchase & transfer fees to nameservice contract.

If all of these look good, i'll go ahead and clean up the integration tests (currently the integration tests have not been updated at all).

I wasn't exactly sure what logic we wanted in fee checking for a few edge case:
- if a string coin value is not parseable to u128 (currently this only result in parse error if the required coin is unparseable, otherwise it returns Insufficient funds error)
- if a required coin value is "0" (currently this will be treated the same as if there was no required coin value set, any amount of sent funds will be valid)

The `coin_helpers.rs` file has some tests which illustrate most of the desired behavior of `assert_sent_sufficient_coin()`.

Any feedback on new contract behavior, helpers, or test coverage is welcome.

I wanted to add the Config state to the examples/schema.rs, but realized that the schema alone wasn't sufficient for exposing an API. Any code trying to interface with raw state would also need to know the prefix of the key, and serialization of the key in the storage layer. Or is it simply expected that someone can gather this by reading the source code?

If we have a standardized way we want to provide this information, I could see us adding a few metadata keys at the beginning of each JSON schema stating a type of schema (Message, State, Response), and what the storage prefix is if it's a state schema.

### DIFF
--- a/nameservice/src/coin_helpers.rs
+++ b/nameservice/src/coin_helpers.rs
@@ -72,3 +72,33 @@ fn assert_sent_sufficient_coin_works() {
         Err(e) => panic!("Unexpected error: {:?}", e),
     };
 }
+
+#[test]
+fn assert_sent_sufficient_coin_handles_parse_failure() {
+    match assert_sent_sufficient_coin(&None, coin("ff", "token")) {
+        Ok(()) => panic!("Should have raised parse error"),
+        Err(Error::ContractErr { msg, .. }) => assert_eq!(msg, "Error while parsing string to u128"),
+        Err(e) => panic!("Unexpected error: {:?}", e),
+    };
+
+    let sent_coins = Some(vec![
+        coin("abcd", "smokin"),
+        coin("5", "token"),
+    ]);
+
+    match assert_sent_sufficient_coin(&sent_coins, coin("5", "token")) {
+        Ok(()) => {}
+        Err(e) => panic!("Unexpected error: {:?}", e),
+    };
+
+    let sent_coins = Some(vec![
+        coin("abcd", "smokin"),
+        coin("efg", "token"),
+    ]);
+
+    match assert_sent_sufficient_coin(&sent_coins, coin("5", "token")) {
+        Ok(()) => panic!("Should have raised parse error"),
+        Err(Error::ContractErr { msg, .. }) => assert_eq!(msg, "Insufficient funds sent"),
+        Err(e) => panic!("Unexpected error: {:?}", e),
+    };
+}
--- a/nameservice/src/coin_helpers.rs
+++ b/nameservice/src/coin_helpers.rs
@@ -1,0 +1,74 @@
+use cosmwasm::errors::{contract_err, Error, Result};
+pub use cosmwasm::types::coin as coin_vec;
+use cosmwasm::types::Coin;
+
+fn parse_u128(source: &str) -> Result<u128> {
+    match source.parse::<u128>() {
+        Ok(value) => Ok(value),
+        Err(_) => contract_err("Error while parsing string to u128"),
+    }
+}
+
+pub fn assert_sent_sufficient_coin(sent: &Option<Vec<Coin>>, required: Coin) -> Result<()> {
+    let required_amount = parse_u128(&required.amount)?;
+
+    match sent {
+        Some(coins) => {
+            if coins.iter().any(|coin| {
+                let amount = parse_u128(&coin.amount).unwrap_or(0);
+                coin.denom == required.denom && amount >= required_amount
+            }) {
+                return Ok(());
+            }
+        }
+        None => {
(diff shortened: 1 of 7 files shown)


ethanfrey:
> I wanted to add the Config state to the examples/schema.rs, but realized that the schema alone wasn't sufficient for exposing an API. Any code trying to interface with raw state would also need to know the prefix of the key, and serialization of the key in the storage layer. Or is it simply expected that someone can gather this by reading the source code?

> If we have a standardized way we want to provide this information, I could see us adding a few metadata keys at the beginning of each JSON schema stating a type of schema (Message, State, Response), and what the storage prefix is if it's a state schema.

That is a good point. And important one to resolve if we think supporting raw queries has any importance. At least documenting the key spaces. I don't have many ideas here. Maybe @webmaster128 has some thoughts on this issue?

nameservice/src/coin_helpers.rs
```diff
@@ -0,0 +1,74 @@
...
+        Some(coins) => {
+            if coins.iter().any(|coin| {
+                let amount = parse_u128(&coin.amount).unwrap_or(0);
+                coin.denom == required.denom && amount >= required_amount
+            }) {
+                return Ok(());
+            }
+        }
+        None => {
+            if required_amount == 0 {
```
ethanfrey: I would do this check first, before the match. We don't need to iterate over all sent coins if there is no fee. The the match can be `if let Some(coins) = sent`...

nameservice/src/coin_helpers.rs
```diff
@@ -0,0 +1,74 @@
...
+
+pub fn coin(amount: &str, denom: &str) -> Coin {
+    Coin {
+        amount: amount.to_string(),
+        denom: denom.to_string(),
+    }
+}
+
+#[test]
+fn assert_sent_sufficient_coin_works() {
```
ethanfrey: nice tests

nameservice/src/contract.rs
```diff
@@ -26,13 +34,18 @@ pub fn handle<S: Storage, A: Api>(
...
 
+
 pub fn try_register<S: Storage, A: Api>(
     deps: &mut Extern<S, A>,
     params: Params,
     name: String,
 ) -> Result<Response> {
-    let key = name.as_bytes();
+    let config_state = config(&mut deps.storage).load()?;
+    if let Some(coin_required) = config_state.purchase_price {
```
ethanfrey: Maybe do this check inside `assert_sent_sufficient_coin`? (always pass if coin_required is `None` or `0`)
That would make the business logic a bit cleaner

nameservice/src/tests.rs
```diff
@@ -0,0 +1,317 @@
...
+
+    match res {
+        Ok(_) => panic!("Must return error"),
+        Err(Error::ContractErr { msg, .. }) => assert_eq!(msg, "Name is already taken"),
+        Err(e) => panic!("Unexpected error: {:?}", e),
+    }
+}
+
+#[test]
+fn fails_on_register_insufficient_fees() {
```
ethanfrey: Nice set of test cases.
I feel that this covers all branches.

I was wondering if there is any simple tool in rust to show code coverage from the tests? (I am sure there is something), so it can be easily clear if tests cover all branches of the contract execution. Meaning we really just need to look carefully at the conditions of the if statements.

ethanfrey:
> I wasn't exactly sure what logic we wanted in fee checking for a few edge case:
>
>    if a string coin value is not parseable to u128 (currently this only result in parse error if the required coin is unparseable, otherwise it returns Insufficient funds error)
>    if a required coin value is "0" (currently this will be treated the same as if there was no required coin value set, any amount of sent funds will be valid)

These both seem very reasonable assumptions

nameservice/src/tests.rs
clevinson: I haven't worked with any rust code coverage tools before. Tried playing around with a few today, and [tarpaulin](https://github.com/xd009642/tarpaulin) seems like one of the better options, and also has some example travis-ci configurations in the readme.

ethanfrey: Looks interesting. Thanks

nameservice/src/coin_helpers.rs
```diff
@@ -9,25 +9,24 @@ fn parse_u128(source: &str) -> Result<u128> {
...
-                coin.denom == required.denom && amount >= required_amount
-            }) {
-                return Ok(());
-            }
-        }
-        None => {
-            if required_amount == 0 {
-                return Ok(());
+pub fn assert_sent_sufficient_coin(sent: &Option<Vec<Coin>>, required: Option<Coin>) -> Result<()> {
+    if let Some(required_coin) = required {
```
ethanfrey: :+1:

PR #35, opened 2020-02-07
Fix cargo caching

webmaster128:
Cache the whole local ./target folder. The wasm build writes a lot of output to ./target/release.


webmaster128:
Speedup of escrow rebuild without change from 02:14 to 00:57

PR #37, opened 2020-02-12
Simplify erc20 contract: Part 1

ethanfrey:
ERC20 inspired many helpers in `cw_storage` (PrefixedStorage was basically copied from there), and error helpers in cosmwasm. Let's import these and use them to simplify the code (and reduce line count).

I also moved the files to the top level, but kept `contract.rs` and `test.rs` separate.

This is byte-wise compatible with the current contract, just cleaned up using newer helper functions. I will make another PR with a deeper change in Storage to use more helpers, but which will change the raw disk storage (but not the public APIs)


erc20/Cargo.toml
```diff
@@ -30,8 +30,11 @@ cranelift = [ "cosmwasm-vm/default-cranelift"]
 singlepass = [ "cosmwasm-vm/default-singlepass"]
 
 [dependencies]
-cosmwasm = { version = "0.6.3" }
+cosmwasm = { version = "0.6.4" }
+cw-storage = "0.1.1"
 schemars = "=0.5"
+named_type = "=0.2.1"
+named_type_derive = "=0.2.1"
```
webmaster128: Why those version fixes with `=`?  Don't we want to receive patched when upgrading dependencies?

erc20/src/lib.rs
```diff
@@ -1,5 +1,8 @@
 pub mod contract;
 
+#[cfg(test)]
+mod tests;
```
webmaster128: Ah, yeah, same effect as I originally intended but simpler file structure 👍

erc20/Cargo.toml
ethanfrey: It hadn't been updated in a year, so I figured safer to pin than allow rogue upgrades.

I just checked at it was bumped to 0.2.2... but just to add a deprecation warning. Seems like this is now in rust std:

https://doc.rust-lang.org/std/any/fn.type_name.html

ethanfrey: Made an issue to fix this up: https://github.com/confio/cw-storage/issues/9

PR #38, opened 2020-02-12
Erc20 storage helpers

ethanfrey:
No longer do manual byte encoding of u128 to store on disk, but use std serde-json compatible types (Amount), so we can use `cw_storage` helpers for all read/write boilerplate

### DIFF
--- a/erc20/src/state.rs
+++ b/erc20/src/state.rs
@@ -1,0 +1,20 @@
+use named_type::NamedType;
+use named_type_derive::NamedType;
+use schemars::JsonSchema;
+use serde::{Deserialize, Serialize};
+
+//use cw_storage::{serialize, PrefixedStorage, ReadonlyPrefixedStorage};
+
+pub const PREFIX_CONFIG: &[u8] = b"config";
+pub const PREFIX_BALANCES: &[u8] = b"balances";
+pub const PREFIX_ALLOWANCES: &[u8] = b"allowances";
+
+pub const KEY_CONSTANTS: &[u8] = b"constants";
+pub const KEY_TOTAL_SUPPLY: &[u8] = b"total_supply";
+
+#[derive(Serialize, Debug, Deserialize, Clone, PartialEq, JsonSchema, NamedType)]
+pub struct Constants {
+    pub name: String,
+    pub symbol: String,
+    pub decimals: u8,
+}
--- a/erc20/src/state.rs
+++ b/erc20/src/state.rs
@@ -3,6 +3,7 @@ use named_type_derive::NamedType;
 use schemars::JsonSchema;
 use serde::{Deserialize, Serialize};
 
+use cosmwasm::errors::{contract_err, Result};
 //use cw_storage::{serialize, PrefixedStorage, ReadonlyPrefixedStorage};
 
 pub const PREFIX_CONFIG: &[u8] = b"config";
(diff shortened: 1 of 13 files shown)


erc20/src/state.rs
```diff
@@ -18,3 +19,118 @@ pub struct Constants {
     pub symbol: String,
     pub decimals: u8,
 }
+
+#[derive(Serialize, Debug, Deserialize, Clone, PartialEq, JsonSchema, NamedType)]
+/// Source must be a decadic integer >= 0
+pub struct Amount(String);
```
webmaster128: I guess we can't do `pub struct Amount(u128)` because of JSON, right?

erc20/src/state.rs
```diff
@@ -18,3 +19,118 @@ pub struct Constants {
...
+
+impl From<u128> for Amount {
+    fn from(val: u128) -> Self {
+        Amount(val.to_string())
+    }
+}
+
+impl From<&str> for Amount {
+    fn from(raw: &str) -> Self {
+        Amount(raw.to_string())
```
webmaster128: This construction from untrusted source should be validated I think

erc20/tests/integration.rs
```diff
@@ -219,7 +220,7 @@ fn transfer_from_works() {
     let transfer_from_msg = HandleMsg::TransferFrom {
         owner: owner.clone(),
         recipient: recipient.clone(),
-        amount: "2".to_string(),
+        amount: Amount::from("2"),
```
webmaster128: Nice syntax

erc20/src/state.rs
ethanfrey: Exactly, we use String for proper JSON encoding and then just parse it when needed

erc20/tests/integration.rs
ethanfrey: Thanks. But with the above comment, I must do something like: 
`Amount::try_from("2").unwrap()`

erc20/src/state.rs
ethanfrey: Note that we cannot control what comes in JSON, so we cannot assume `Amount(String)` is always valid string. We must do `.validate()` or `.parse()` to validate. This is actually the same as the original, that parsed input strings.

Also, this is designed for test cases, and we do want to run tests with invalid data. So, I think allowing bad data via this constructor is okay (but all business logic must check it - cannot take Amount in from user and store to disk)

erc20/tests/integration.rs
webmaster128: Well, Rust enforces the truth. Life is more complex than we wish

erc20/src/state.rs
webmaster128: Hmm, in this case the existence of a dedicated `Amount` is misleading. If there is such a type, I expect that it holds valid data. Having an explicit conversion from untrusted souce is fine, but then untrusted data should remain in a String type.

ethanfrey: We do allow user-defined JSON to hold a possibly invalid `HumanAddr`.
But I agree, anything that has been validated from users must correct (once it is in the system)

ethanfrey:
Okay, there are a few failing tests and I found some limits to `cw-storage`.
I want to update `cw-storage`, then come back to this.

I think the business logic in `contract` is cleaner.
I pulled the String -> u128 and overflow check logic into state, and the message validation logic into validate.
Otherwise, heavily using cw-storage helpers to handle all the reads and writes.

I think the main file looks cleaner, and the checks are clearly specified. But that's just my opinion.
I will get back to this in a day or two

ethanfrey:
This depends on https://github.com/confio/cw-storage/pull/10 (or some modification of that) being merged and tagged.

ethanfrey:
After some reflection, and noticing the loc doesn't really change, I wonder if this is worth it.

It does provide higher level abstractions around amount and data storage and I find the business logic easier to read. But is it better? I could be convinced to undo most of this.

One simple piece that I do like and can be separated is adding validate logic to the messages. So we run that before even starting to process them, which seems like a good way to avoid letting unchecked data slip though some conditional branch

ethanfrey:
I think to close this and maybe revisit this idea later.

Unless you think there is anything here that should be extracted and merged @webmaster128

webmaster128:
I looked though it again. There might be some improvements in here, but there are other things I dislike, especially additional types that need to be validated manually and don't provide type safety automatically. I agree with closing it and if something in here should still be done, then I guess it is better to break it down into smaller, independent steps.

PR #41, opened 2020-02-24
update examples to use msg.rs

marbar3778:
- currently in the template there is a msg dir while in some examples there is not. While this is a uber minor thing it makes it a bit nicer to have it all uniform

Signed-off-by: Marko Baricevic <marbar3778@yahoo.com>

### DIFF
--- a/erc20/examples/schema.rs
+++ b/erc20/examples/schema.rs
@@ -4,9 +4,8 @@ use std::path::PathBuf;
 
 use schemars::{schema::RootSchema, schema_for};
 
-use cw_erc20::contract::{
-    AllowanceResponse, BalanceResponse, Constants, HandleMsg, InitMsg, QueryMsg,
-};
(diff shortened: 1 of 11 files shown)


ethanfrey:
Thanks for the initiative.

I actually have a big cleanup of messages in another PR as well: https://github.com/confio/cosmwasm-examples/pull/38/files

If that one doesn't get merged for being too ambitious, I would merge this. Or see how to combine them somehow. What do you think of this @webmaster128 ?

ethanfrey:
I agree with moving the query responses into the same file as QueryMsg.

In the template (and as an example moving forward), I had one file for msg (public interface), and one for state (how we manage internal store), so separate out all structs

ethanfrey:
Seems good, I'll merge now, and deal with conflicts with my pending PR later. This is a subset of what I want to do, but a good subset.

Issue #42, opened 2020-02-26
Upgrade all contracts to cosmwasm 0.7


ethanfrey:
This involves adjusting to all breaking changes. But also making use of enhancements, like Base64/Binary for better representation of binary data in json (see #39)

webmaster128:
I'm working on the erc20 now

PR #45, opened 2020-02-27
Migrate cw-erc20 to cosmwasm 0.7.0

webmaster128:
Part of #42


erc20/src/contract.rs
```diff
@@ -123,20 +121,27 @@ fn try_transfer<S: Storage, A: Api>(
 
     let res = Response {
         messages: vec![],
-        log: Some("transfer successful".to_string()),
+        log: vec![
```
ethanfrey: nicer. hope this works well in the js interface.

PR #47, opened 2020-02-27
Upgrade nameservice contract to cosmwasm 0.7.0

### LINKED ISSUE #42: Upgrade all contracts to cosmwasm 0.7

webmaster128:
Version 0.1.0 was published to https://crates.io/crates/cw-nameservice. I'll publish 0.2.0 as soon as this is merged.

Closes #42


nameservice/src/contract.rs
```diff
@@ -61,23 +61,27 @@ pub fn try_register<S: Storage, A: Api>(
...
+            if env.message.signer != record.owner {
+                unauthorized()?;
+            }
 
-        record.owner = new_owner.clone();
-        Ok(record)
+            record.owner = new_owner.clone();
+            Ok(record)
+        } else {
+            unauthorized()
```
ethanfrey: nitpick: I would return not_found here (difference from not being able to transfer someone else's name, and the name not being registered at all)

ethanfrey: (This was also the old behavior, implicit when calling `load` not `may_load` inside `cw-storage`)

webmaster128: Thanks, will update! I was not sure there and did the migration somewhat mechanically.

webmaster128: NotFound does not have a helper function and I think the `kind` field with a type is not very helpful. I used a `contract_err` now (74d79f5e9c4d21c53f4367bacc46f7b5a21a7564)

webmaster128:
I'm rebuilding the contract.wasm right now

ethanfrey:
Closes #42 right?

webmaster128:
Right. Done from my side. Good to merge?

PR #50, opened 2020-03-05
Remove wasm-bindgen from all contracts

webmaster128:
- Remove wasm-bindgen from all contracts
- Exclude contract.wasm and hash.txt from source code publication
- Update the 3 docs 1:1 from the template repo. I think all desired changes should be done there since 1:1 copy is the only way we can keep them mostly up to date if we want them in every example.

### DIFF
--- a/erc20/Cargo.lock
+++ b/erc20/Cargo.lock
@@ -97,11 +97,6 @@ dependencies = [
 ]
 
 [[package]]
-name = "bumpalo"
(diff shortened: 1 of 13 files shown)


webmaster128:
> And yes, maybe we should have a script to update the md files from templates (or maybe you do have one locally)

I don't have one yet. As long as we keep the files generic enough to do just `cp`, that's no problem.

PR #55, opened 2020-03-17
Nameservice fixes

### LINKED ISSUE #52: Make non-existent name a valid query result
When you query a name that does not exist, you get the error

> GET https://lcd.demo-07.cosmwasm.com/wasm/contract/cosmos1vjecguu37pmd577339wrdp208ddzymkudc46zj/smart/7b227265736f6c76657265636f7264223a7b226e616d65223a2273696d6f6e227d7d?encoding=hex 500 (Internal Server Error)
>
> Error: query wasm contract failed: cw_nameservice::state::NameRecord not found (HTTP 500)
    at parseAxiosError (restclient.ts:232)
    at async RestClient.get (restclient.ts:254)
    at async RestClient.queryContractSmart (restclient.ts:409)

This is in any case not a server error (http 5xx). I also think a missing name should not be treated as an error. Instead, `QueryMsg::ResolveRecord` should have some kind of optional return type.

ethanfrey:
Closes #52 and then some

- [x] Enforces schema on name registration: `[a-z0-9._\-]{3,64}`
- [x]  Returns `None` on queries for non-regitered name (rather than error)


nameservice/src/contract.rs
```diff
@@ -102,3 +108,31 @@ fn query_resolver<S: Storage, A: Api>(deps: &Extern<S, A>, name: String) -> Resu
 
     serialize(&resp)
 }
+
+// let's not import a regexp library and just do these checks by hand
+fn invalid_char(c: &char) -> bool {
+    if c >= &'0' && c <= &'9' {
```
webmaster128: This performs numeric comparisons on the reference. Is this what we want? In C this woud compare pointers instead of values. Even if legal, I think it is better to dereference `c`.

nameservice/src/contract.rs
```diff
@@ -95,10 +100,40 @@ pub fn query<S: Storage, A: Api>(deps: &Extern<S, A>, msg: QueryMsg) -> Result<V
...
+        false
+    } else if c == &'.' || c == &'-' || c == &'_' {
+        false
+    } else {
+        true
+    }
+}
+
+/// validate_name returns an error if the name is invalid
+/// (we require 4-64 lowercase ascii letters, numbers, or . - _)
```
webmaster128: 3-64

nameservice/tests/integration.rs
```diff
@@ -347,10 +345,8 @@ fn fails_on_query_unregistered_name() {
...
-    );
-
-    match res {
-        QueryResult::Ok(_) => panic!("Must return error"),
-        QueryResult::Err(e) => assert_eq!(e, "cw_nameservice::state::NameRecord not found"),
-    }
+    )
+    .unwrap();
+    let value: ResolveRecordResponse = deserialize(res.as_slice()).unwrap();
+    assert_eq!(value.address.as_str(), "");
```
webmaster128: We sould adapt the test name to the new expected behaviour

nameservice/src/contract.rs
```diff
@@ -95,10 +100,40 @@ pub fn query<S: Storage, A: Api>(deps: &Extern<S, A>, msg: QueryMsg) -> Result<V
 fn query_resolver<S: Storage, A: Api>(deps: &Extern<S, A>, name: String) -> Result<Vec<u8>> {
     let key = name.as_bytes();
 
-    let record = resolver_read(&deps.storage).load(key)?;
-    let address = deps.api.human_address(&record.owner)?;
+    let address = match resolver_read(&deps.storage).may_load(key)? {
+        Some(record) => deps.api.human_address(&record.owner)?,
+        None => HumanAddr::from(""),
+    };
```
webmaster128: Would it be possible to express this case as

```ts
export interface ResolveRecordResponse {
  address: string | null;
}
```

?

ethanfrey: I will try

nameservice/src/contract.rs
ethanfrey: This does work. I will try deref. This was the suggestion from the compiler.

nameservice/src/contract.rs
ethanfrey: It doesn't return null, but I made it optional. So we get:

```typescript
export interface ResolveRecordResponse {
  address: string | undefined;
}
```

This should be close to what we wanted

webmaster128: Much better for sure. Since there is no `undefined` in JSON, it will result in `{}`, so the interface should be

```ts
export interface ResolveRecordResponse {
  address?: string;
}
```

nameservice/schema/resolve_record_response.json
```diff
@@ -7,7 +7,14 @@
   ],
   "properties": {
     "address": {
-      "$ref": "#/definitions/HumanAddr"
+      "anyOf": [
+        {
+          "$ref": "#/definitions/HumanAddr"
+        },
+        {
+          "type": "null"
```
webmaster128: Are you sure you get undefined at JSON level?

ethanfrey: You are right, it outputs: `{"address":null}`

ethanfrey:
I think I addressed all issues.

nameservice/src/contract.rs
```diff
@@ -95,10 +100,40 @@ pub fn query<S: Storage, A: Api>(deps: &Extern<S, A>, msg: QueryMsg) -> Result<V
...
+        None => None,
+    };
     let resp = ResolveRecordResponse { address };
 
     serialize(&resp)
 }
+
+// let's not import a regexp library and just do these checks by hand
+fn invalid_char(ch: &char) -> bool {
+    let c = *ch;
```
webmaster128: ❤️

PR #56, opened 2020-03-17
Add Burn msg to erc20 contract

anilCSE:


erc20/src/contract.rs
```diff
@@ -211,6 +212,57 @@ fn try_approve<S: Storage, A: Api>(
     Ok(res)
 }
 
+/**
+* Destroy tokens
+*
+* Remove `amount` tokens from the system irreversibly
+*
+* @param amount the amount of money to burn
+*/
```
webmaster128: Rust uses triple slash doc comments, like

```rust
/// Destroy tokens
///
/// Remove `amount` tokens from the system irreversibly
...
```

webmaster128: I'd mention who's balance is reduced by the burning

anilCSE:
> Very nice, I like. Let @ethanfrey have a look as well.
> 
> Can you rebase on top of current master? ` erc20/contract.wasm` and `erc20/hash.txt` were updated recently on master, which is causing the conflicts.

Updated! Thank you for a quick review.

webmaster128:
Right now the CI fails because of a missing `cargo fmt && cargo schema`

anilCSE:
> Right now the CI fails because of a missing `cargo fmt && cargo schema`

Updated

erc20/src/contract.rs
```diff
@@ -211,6 +212,57 @@ fn try_approve<S: Storage, A: Api>(
...
+    if account_balance < amount_raw {
+        return dyn_contract_err(format!(
+            "insufficient funds to burn: balance={}, required={}",
+            account_balance, amount_raw
+        ));
+    }
+    account_balance -= amount_raw;
+    balances_store.set(owner_address_raw.as_slice(), &account_balance.to_be_bytes());
+
+    let mut config_store = PrefixedStorage::new(PREFIX_CONFIG, &mut deps.storage);
```
ethanfrey: Good, update total supply as well.

erc20/src/tests.rs
```diff
@@ -1033,6 +1033,167 @@ mod transfer_from {
...
+        ); // -1
+        assert_eq!(
+            get_balance(&deps.api, &deps.storage, &HumanAddr("addr1111".to_string())),
+            22
+        );
+        assert_eq!(get_total_supply(&deps.storage), 32);
+    }
+
+    #[test]
+    fn can_burn_zero_amount() {
```
ethanfrey: I would add a test of burning from an account with no balance. (should error)

anilCSE: I think `fails_on_insufficient_balance` covers that right? I can add one but looks duplicate

anilCSE: May be we can instead have validation for `amount` and it should be > 0

webmaster128: Hmm... I tend to agree with @anilCSE. However, you could treat those cases as different things:
- a user of the contract has insufficient balance, but is known to the contract
- an address that was never seen by the contract before wants to burn

However, this woud test _get balance of address_ more than the burn.

erc20/src/contract.rs
```diff
@@ -211,6 +212,57 @@ fn try_approve<S: Storage, A: Api>(
...
+fn try_burn<S: Storage, A: Api>(
+    deps: &mut Extern<S, A>,
+    env: Env,
+    amount: &str,
+) -> Result<Response> {
+    let owner_address_raw = &env.message.signer;
+    let amount_raw = parse_u128(amount)?;
+
+    let mut balances_store = PrefixedStorage::new(PREFIX_BALANCES, &mut deps.storage);
+    let mut account_balance = read_u128(&balances_store, owner_address_raw.as_slice())?;
```
webmaster128: Wouldn't it make sense to use `read_balance` here? Then we could unit test this to ensure a non existent addresses leads to 0.

erc20/src/tests.rs
webmaster128: > May be we can instead have validation for `amount` and it should be > 0.
> Burning zero tokens doesn't make sense. It's unnecessary gas cost, we should just throw error there

I would avoid adding unnecessary code branches. There is no requirement to disallow burning 0. You can send 0 as well.

anilCSE: Agreed

erc20/src/contract.rs
anilCSE: Why `read_balance` is returning `Result<u128> `and why not just `u128`? I believe read_balance should handle the errors and return just `0`. If we need to validate account address as well, that should be done outside of `read_balance` imo.

erc20/src/contract.rs
```diff
@@ -225,8 +225,7 @@ fn try_burn<S: Storage, A: Api>(
     let owner_address_raw = &env.message.signer;
     let amount_raw = parse_u128(amount)?;
 
-    let mut balances_store = PrefixedStorage::new(PREFIX_BALANCES, &mut deps.storage);
-    let mut account_balance = read_u128(&balances_store, owner_address_raw.as_slice())?;
+    let mut account_balance = read_balance(&mut deps.storage, owner_address_raw).unwrap();
```
ethanfrey: Don't unwrap, return an error.

Replace `.unwrap()` with `?`. Works like magic.

PR #59, opened 2020-03-17
Add mask contract

### LINKED ISSUE #24: Add mask/reflector contract
The idea is a simple demo contract, which can be used to help test https://github.com/confio/cosmwasm/issues/90 and https://github.com/cosmwasm/wasmd/issues/16 as well as serve as a demo.

This contract is created with an owner, and has two actions:
* `ReflectMsg` will take a `CosmosMsg` as an argument, and if it was signed by the owner, it will re-dispatch the same message untouched. This let's the owner use the "mask" as his identity for anything - send tokens, stake, etc.
* `ChangeOwner` will take a `HumanAddr` as an argument. It must be signed by the current owner, and will update to a new owner.

This will let us easily test multiple scenarios in re-dispatching and handling "opaque messages". However, it is actually a useful functionality in some cases. Imagine "validator key rotation". If you create a validator with such a "mask contract" as the owner, you can swap out control of the mask without updating the validator struct at all. A mask could also own staked tokens, and you could make a trade of the mask against some liquid tokens if desired. (Since cosmos-sdk allows multiple messages in one tx, you can do it atomically by doing a multisig off-chain, then executing both transfers together)

ethanfrey:
Closes #24 

* Fresh contract generated from cosmwasm-template
* Pulls in logic code from quick mask demo
* Adds full test coverage


mask/src/state.rs
```diff
@@ -0,0 +1,21 @@
+use schemars::JsonSchema;
+use serde::{Deserialize, Serialize};
+
+use cosmwasm::traits::Storage;
+use cosmwasm::types::CanonicalAddr;
+use cw_storage::{singleton, singleton_read, ReadonlySingleton, Singleton};
+
+pub static CONFIG_KEY: &[u8] = b"config";
```
webmaster128: When you have those wrappers below, can this be privatized?

mask/src/msg.rs
```diff
@@ -0,0 +1,26 @@
...
+
+use cosmwasm::types::{CosmosMsg, HumanAddr};
+
+#[derive(Serialize, Deserialize, Clone, Debug, PartialEq, JsonSchema)]
+pub struct InitMsg {}
+
+#[derive(Serialize, Deserialize, Clone, Debug, PartialEq, JsonSchema)]
+#[serde(rename_all = "lowercase")]
+pub enum HandleMsg {
+    ReflectMsg { msg: CosmosMsg },
```
webmaster128: How do you feel about making this an array of CosmosMsg?

mask/src/msg.rs
```diff
@@ -0,0 +1,26 @@
...
+    ChangeOwner { owner: HumanAddr },
+}
+
+#[derive(Serialize, Deserialize, Clone, Debug, PartialEq, JsonSchema)]
+#[serde(rename_all = "lowercase")]
+pub enum QueryMsg {
+    GetOwner {},
```
webmaster128: I'm thinking if we should start all of those with a verb or the the subject that is queried. In the nameservice contract it is mixed:

```rust
pub enum QueryMsg {
    // ResolveAddress returns the current address that the name resolves to
    ResolveRecord { name: String },
    Config {},
}
```

I have a slight tendency for using the subject only, as it will reduce noise a bit. But I'm unsure.

ethanfrey: Okay, I agree.. let's just do subject. One word is better for message readability

mask/src/msg.rs
ethanfrey: sure, why not.
I would enforce len > 0

mask/src/state.rs
ethanfrey: yup

ethanfrey:
@webmaster128 all change requests addressed, ready for review

PR #63, opened 2020-04-08
Update escrow expiry types

webmaster128:
This

1. Encodes setting `end_height`/`end_type` as Option instead of zero value
2. Improves documentation of the fields, clarifying that each of them can independently trigger an expiery
3. Fix `>=` to `>` in `is_expired` to match the documented behaviour

I'd also love to allow the creation of expired escrows, but I'm afraid this starts a long debate about minimalism and correctness vs. preventing users to shoot themselves into the foot.

### DIFF
--- a/escrow/src/contract.rs
+++ b/escrow/src/contract.rs
@@ -19,13 +19,13 @@ pub struct State {
 impl State {
     fn is_expired(&self, env: &Env) -> bool {
         if let Some(end_height) = self.end_height {
-            if env.block.height >= end_height {
+            if env.block.height > end_height {
                 return true;
             }
         }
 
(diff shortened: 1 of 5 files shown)


escrow/src/contract.rs
```diff
@@ -12,14 +12,25 @@ pub struct State {
     pub arbiter: CanonicalAddr,
     pub recipient: CanonicalAddr,
     pub source: CanonicalAddr,
-    pub end_height: i64,
-    pub end_time: i64,
+    pub end_height: Option<i64>,
```
ethanfrey, line 15: Yes, this is good.

Issue #65, opened 2020-05-15
Update to CosmWasm 0.8


webmaster128:
- [x] erc20 #68
- [x] escrow #72
- [x] mask #69
- [x] nameservice #70

webmaster128:
Done

PR #68, opened 2020-05-16
Update erc20 contract to 0.8

webmaster128:
Based on #67

Part of #65

### DIFF
--- a/erc20/src/contract.rs
+++ b/erc20/src/contract.rs
@@ -3,10 +3,11 @@ use serde::{Deserialize, Serialize};
 use std::convert::TryInto;
 
 use crate::msg::{AllowanceResponse, BalanceResponse, HandleMsg, InitMsg, QueryMsg};
-use cosmwasm::errors::{contract_err, dyn_contract_err, Result};
-use cosmwasm::traits::{Api, Extern, ReadonlyStorage, Storage};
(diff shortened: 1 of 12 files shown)


erc20/src/contract.rs
```diff
@@ -133,46 +141,46 @@ fn try_transfer<S: Storage, A: Api>(
...
 
-    let res = Response {
+    let res = HandleResponse {
         messages: vec![],
         log: vec![
             log("action", "transfer_from"),
             log(
                 "spender",
-                deps.api.human_address(&env.message.signer)?.as_str(),
+                deps.api.human_address(&env.message.sender)?.as_str(),
```
ethanfrey, line 183: We were discussing to remove the human / canonical address division if we use a binary encoding scheme, so we can just pass in canonical addresses. 

This is one place where we will want to convert to human readable in any case. So we will need the hooks, just have to use them much less

webmaster128:
> But if you are open, I can suggest a much more radical change using Uint128 for input types and making full use of cosmwasm storage.

Updated the public interfaces to Uin128 – good point. One helper function less we need. However, I kept the balance and allowance storage 16 raw bytes, which is more compact that using JSON encoded decimal strings. We can change that later if we want.

ethanfrey:
Fair enough, let's merge this!

PR #69, opened 2020-05-16
Update mask contract to 0.8

webmaster128:
Based on #67

Part of #65


ethanfrey:
What is blocking this btw?

After thinking some more I would not deprecate this, but rather I would extend mask to make it more of a proxy, 1 of N "multisig" in the future (by 0.9)

webmaster128:
This just needs an 0.8.0-rc1 -> 0.8.0 update

webmaster128:
> I would (after 0.8) consider either deprecating this can just leaving the new reflect contract for debug demos, or expand this to offer some real functionality beyond a developer swiss army knife

I we need this example or not, I don't know. But I think it is important to not mix up development contracts and example contracts. The later ones are needed at higher level for apps, explorers and build verification. They have proper versioning and a published source code.

PR #72, opened 2020-05-16
Update escrow contract to 0.8

webmaster128:
Based on #67

Part of #65

---------

There are a few TODOs left.

~Biggest question is how to update mock balances in integration tests. I found a way for unit tests but this cannot be adapted directly.~

### DIFF
--- a/escrow/src/contract.rs
+++ b/escrow/src/contract.rs
@@ -54,10 +54,9 @@ fn try_approve<S: Storage, A: Api, Q: Querier>(
         } else {
             // release everything
 
-            // TODO: avoid the necessarity of this conversion
-            let human = deps.api.human_address(&env.contract.address)?;
-            // TODO: make sure querier guarantees to returns up-to-date data
-            deps.querier.query_all_balances(human)?
+            let contract_address_human = deps.api.human_address(&env.contract.address)?;
+            // TODO: Ensure querier guarantees to returns up-to-date data. What about funds sent in this handle message?
+            deps.querier.query_all_balances(contract_address_human)?
         };
 
         send_tokens(
@@ -79,10 +78,9 @@ fn try_refund<S: Storage, A: Api, Q: Querier>(
     if !state.is_expired(&env) {
         Err(generic_err("escrow not yet expired"))
     } else {
-        // TODO: avoid the necessarity of this conversion
-        let human = deps.api.human_address(&env.contract.address)?;
-        // TODO: make sure querier guarantees to returns up-to-date data
-        let balance = deps.querier.query_all_balances(human)?;
+        let contract_address_human = deps.api.human_address(&env.contract.address)?;
+        // TODO: Ensure querier guarantees to returns up-to-date data. What about funds sent in this handle message?
+        let balance = deps.querier.query_all_balances(contract_address_human)?;
         send_tokens(
             &deps.api,
             &env.contract.address,
--- a/escrow/src/contract.rs
+++ b/escrow/src/contract.rs
@@ -2,42 +2,9 @@ use cosmwasm_std::{
     generic_err, log, unauthorized, Api, BankMsg, Binary, CanonicalAddr, Coin, CosmosMsg, Env,
     Extern, HandleResponse, HandleResult, InitResponse, InitResult, Querier, StdResult, Storage,
 };
-use cosmwasm_storage::{singleton, Singleton};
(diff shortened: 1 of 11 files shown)


escrow/src/contract.rs
```diff
@@ -1,90 +1,66 @@
...
     } else {
-        #[allow(clippy::or_fun_call)]
-        let amount = quantity.unwrap_or(env.contract.balance.unwrap_or_default());
+        let amount = if let Some(quantity) = quantity {
+            quantity
+        } else {
+            // release everything
+
+            let contract_address_human = deps.api.human_address(&env.contract.address)?;
+            // TODO: Ensure querier guarantees to returns up-to-date data. What about funds sent in this handle message?
```
ethanfrey, line 58: Yup, that is part of the cosmos-sdk integration tests.
Basically we need to define how this should work (include all balance up to those sent with the same message), then some tests on `wasmd` that it does work properly, and the rest of the contracts can depend on it.

ethanfrey: I'm not sure I have such a test (but have other ones on staking state). If you want to make an issue on wasmd to test this and on cosmwasm to document this, then you can remove the TODO (and just trust this works properly)

escrow/src/contract.rs
```diff
@@ -202,175 +185,126 @@ mod tests {
...
+        let init_env = mock_env_height(&deps.api, "creator", &init_amount, 876, 0);
+        let contract_addr = deps.api.human_address(&init_env.contract.address).unwrap();
         let msg = init_msg_expire_by_height(1000);
-        let env = mock_env_height(&deps.api, "creator", &coin("1000", "earth"), &[], 876, 0);
-        let init_res = init(&mut deps, env, msg).unwrap();
+        let init_res = init(&mut deps, init_env, msg).unwrap();
         assert_eq!(0, init_res.messages.len());
 
+        // balance changed in init
+        deps.querier.update_balance(&contract_addr, init_amount);
```
ethanfrey, line 212: If we define that "balance includes all tokens sent in the same message", which seems to be what we want to do, then we can set this balance above the `init` call (maybe right after the `mock_env_height`). This would be closest to the previous style.

Note that the args were wrong on the previous test - the send funds must be included in the balance.

escrow/tests/integration.rs
```diff
@@ -50,201 +42,162 @@ fn init_msg_expire_by_height(height: i64) -> InitMsg {
...
+    let init_env = mock_env_height(&deps.api, "creator", &init_amount, 876, 0);
+    let contract_addr = deps.api.human_address(&init_env.contract.address).unwrap();
     let msg = init_msg_expire_by_height(1000);
-    let env = mock_env_height(&deps.api, "creator", &coin("1000", "earth"), &[], 876, 0);
-    let init_res = init(&mut deps, env, msg).unwrap();
+    let init_res: InitResponse = init(&mut deps, init_env, msg).unwrap();
     assert_eq!(0, init_res.messages.len());
 
+    // balance changed in init
+    deps.with_querier(|querier| {
```
ethanfrey, line 108: It seems like this TODO from the description is fixed now?

> Biggest question is how to update mock balances in integration tests. I found a way for unit tests but this cannot be adapted directly.

webmaster128: Jap, the description is outdated

escrow/src/contract.rs
webmaster128: How did you implement this? Are the tokens sent with this message included in the balance I query?

If yes, great. If no, we should reject HandleMsg::Approve and HandleMsg::Refund messages that include tokens.

ethanfrey: I am 95% sure I implemented it as I described, and 80% sure I have not tested this.

What I said is "how it should be", but we need to document this and test the implementation.

ethanfrey:
I just updated and pushed. Let me know if I can merge.

I want this example, so I can finish updating the docs (it is based on forking this): https://github.com/CosmWasm/docs/pull/45

webmaster128:
Nice. See https://github.com/CosmWasm/cosmwasm-examples/commit/7a4c2b0f2b3ba1e2290c2d4e22bee80ce88aea75 and https://github.com/CosmWasm/cosmwasm-examples/commit/59898e2412e1bf239dd7815f57c79f3132a4b749 on master for what is missing for the tagging

PR #77, opened 2020-06-23
Upgrade to 0.9

webmaster128:
Waiting for the final 0.9.0 tag before merge


escrow/tests/integration.rs
```diff
@@ -62,7 +62,7 @@ fn proper_initialization() {
     let api = deps.api;
     deps.with_storage(|store| {
         let config_key_raw = to_length_prefixed(b"config");
-        let state: State = from_slice(&store.get(&config_key_raw).unwrap().unwrap()).unwrap();
+        let state: State = from_slice(&store.get(&config_key_raw).unwrap().0.unwrap()).unwrap();
```
ethanfrey, line 65: Okay, so this was a bit confusing before and more so now.

Definitely not a normal use-case, but maybe we can imagine a helper for this in the future.

webmaster128: Well, this is a VM<->blockchain interface. Very powerful, but don't expect it to be developer frindly

webmaster128: One thing we could do is using a struct instead of a tuple, which would at at least give us property names instead of `.0` and `.1`

ethanfrey: The tuple is fine. And double unwraps are ugly anyway. And yes, this is designed for VM<->Blockchain not testing.

One idea would be some helper like:

`let state: State = testing::must_get_vm_storage(&store, &config_key_raw);`

and then

```rust
pub fn must_get_vm_storage<S: cosmwasm_vm::Storage, T: Deserialize>(store: &S, key: &[u8]) -> T {
  from_slice(&store.get(key).unwrap().unwrap()).unwrap()
}
```

I mean this helper could be anywhere even in the test file itself. Just my idea on how to make this a bit more usable

PR #79, opened 2020-06-25
Add query to escrow

### LINKED ISSUE #78: Create query for escrow
Right now QueryMsg is an enum with no cases. This results in a JSON schema which triggers a warning:

```
{
  "$schema": "http://json-schema.org/draft-07/schema#",
  "title": "QueryMsg",
  "enum": []
}
```

>  Array has too few items. Expected 1 or more.

When trying to create TS definitions, this leads to a crash:

```
SyntaxError: '=>' expected. (2:1)
  1 | export type QueryMsg = ()
> 2 | 
    | ^
```

We have a query for a similar contract: https://github.com/CosmWasm/cosmwasm/blob/v0.9.0/contracts/hackatom/src/contract.rs#L57-L59

webmaster128:
Closes #78

### DIFF
--- a/escrow/schema/query_msg.json
+++ b/escrow/schema/query_msg.json
@@ -1,5 +1,18 @@
 {
   "$schema": "http://json-schema.org/draft-07/schema#",
   "title": "QueryMsg",
-  "enum": []
+  "anyOf": [
+    {
+      "description": "Returns a human-readable representation of the arbiter.",
+      "type": "object",
+      "required": [
+        "arbiter"
+      ],
+      "properties": {
+        "arbiter": {
+          "type": "object"
+        }
+      }
+    }
+  ]
 }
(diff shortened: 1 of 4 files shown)


escrow/schema/query_msg.json
```diff
@@ -1,5 +1,18 @@
 {
   "$schema": "http://json-schema.org/draft-07/schema#",
   "title": "QueryMsg",
-  "enum": []
```
webmaster128, line 4: FYI: This line removed (enum with 0 cases) is invalid accirdong to the JSON schema stadard

ethanfrey: Did you add a valid query just so the json schema would work?

We will want to allow "Never" type objects somehow, eventually. I would be nice to see if there is a way to express this and make an issue on schemars about it (no rush, but good to start such moving)

webmaster128: > Did you add a valid query just so the json schema would work?

Yes. I think if a contract does not support queries, it should just not have the export (like migrate).

> We will want to allow "Never" type objects somehow, eventually. 

This is open for discission I think. A type that cannot be instantiated is an edge case, that works in Rust where the world is correct and mathematical. But as soon as you leave that, I would not expect mappings to types that are equal.

PR #81, opened 2020-07-02
Add a simple cross-contract calling example

louisliu2048:
Add a simple cross-contract calling example, using the cosmwasm-template@v0.8.0


ethanfrey:
Thank you for the PR, why did you close this?

PR #85, opened 2020-07-25
Update contracts to 0.10

maurolacy:
The migration work to 0.10.0.

Initially I needed to add a `patch`section to `Cargo.toml`:
```
[patch.crates-io]
cosmwasm-std = { path = "../../cosmwasm/packages/std" }
cosmwasm-storage = { path = "../../cosmwasm/packages/storage" }
cosmwasm-vm = { path = "../../cosmwasm/packages/vm" }
cosmwasm-schema = { path = "../../cosmwasm/packages/schema" }
``` 

Didn't commit it, as I understand it's temporary. 

Same with the version, that should be changed to `0.10.0`, when published in crates.io.

Feel free to suggest corrections / improvements.


ethanfrey:
Thank you. Curious why referencing the published alpha version didn't work.

I'll take a look later, likely Monday

maurolacy:
> Thank you. Curious why referencing the published alpha version didn't work.

In fact it ended up working... maybe after `rust update`? I'm not sure. 

> 
> I'll take a look later, likely Monday

Sure! I'll take a break too... Tomorrow I'll probably check / work on the other repos' migrations.
Will also take a look at the CI tests.

escrow/Cargo.toml
```diff
@@ -1,6 +1,6 @@
 [package]
 name = "cw-escrow"
-version = "0.5.2"
+version = "0.5.3"
```
ethanfrey, line 3: Same comment about bumping to 0.6.0

mask/Cargo.toml
```diff
@@ -1,6 +1,6 @@
 [package]
 name = "cw-mask"
-version = "0.3.2"
+version = "0.3.3"
```
ethanfrey, line 3: And 0.4.0

nameservice/Cargo.toml
```diff
@@ -1,7 +1,7 @@
 [package]
 name = "cw-nameservice"
 description = "A name service"
-version = "0.5.2"
+version = "0.5.3"
```
ethanfrey, line 4: And 0.6.0 again

erc20/Cargo.toml
```diff
@@ -1,7 +1,7 @@
 [package]
 name = "cw-erc20"
 description = "An implementation of the ERC20 token interface"
-version = "0.5.3"
+version = "0.6.0"
```
ethanfrey, line 4: Thanks. I think the Cargo.lock files need to get updated as well (just cargo build to do so)

ethanfrey: Haha, you already did this while I was reviewing 😄

maurolacy: Yes. I'm also running the optimizer again, as the `contract.wasm` and `hash.txt` will probably change too.

webmaster128: > Yes. I'm also running the optimizer again, as the `contract.wasm` and `hash.txt` will probably change too.

As far as I know, they don't change when you update the version of the contract

maurolacy: Well, they changed. But I think you're right, as I've checked and the change seems not to be related to the version. That is, I just executed the optimizer again (without changing the version, that is), and `contract.wasm` changed(!).

You might want to take a look at it. I'll commit the new `contract.wasm` and `hash.txt` files for reference.

maurolacy: I wonder if this can be somehow related to docker version changes and/or OS libraries... I'm running this on a Mac. Will check in Linux too.

maurolacy: The same happened on Linux, but ONLY after I updated the repo with the contract version changes. So, it seems there's a version effect after all. If I now do a `git reset` to the old commit (`eae4e5056825e892befd158ef50946e6244f85b3`), and re-run the optimizer, the version change still happens.

ethanfrey: If I understand correctly.

`hash(contract with version 0.5.3) != hash(contract with version 0.6.0)` even if the code doesn't change.

But... did you manage to reproduce that the exact same commit lead to different contract hashes? Even if running once on linux and once on osx? If there is any way to build the identical code into different hashes via `rust-optimizer`, then I want to track it down and remove it. It is supposed to be deterministic.

maurolacy: To be clear, both Linux and OSX produce the same hashes.

What I found out was that the after changing the version, the hashes changed. You can reproduce this by updating your branch, and running the optimizer:
```
git checkout update-contracts-to-0.10
git pull --rebase
cd erc20/
docker run --rm -v "$(pwd)":/code --mount type=volume,source="$(basename "$(pwd)")_cache",target=/code/target --mount type=volume,source=registry_cache,target=/usr/local/cargo/registry cosmwasm/rust-optimizer:0.9.0
```
I'm posting the commands just for reference. You should get the same hash than the committed one.

After that, what I did was:
```
git checkout eae4e5056825e892befd158ef50946e6244f85b3
docker run --rm -v "$(pwd)":/code --mount type=volume,source="$(basename "$(pwd)")_cache",target=/code/target --mount type=volume,source=registry_cache,target=/usr/local/cargo/registry cosmwasm/rust-optimizer:0.9.0
```
and the hash is still the new one(!).

So, there seem to be a caching / version issue going on at the Rust level, that only manifested when I upgraded the contract version.

ethanfrey: Yes, this `--mount type=volume,source="$(basename "$(pwd)")_cache"` is the caching of the partial compilations, which makes it much faster in dev mode.

You can remove this, or better `docker volume rm "$(basename "$(pwd)")_cache"` to wipe it clean and recompile. It will take quite some time, but you should be able to reproduce hashes

maurolacy: Okay, good to know. I'll check later and report back. I'll also always use this variant to generate contract / hashes for reference.

maurolacy: OK, I confirmed this. I would have expected Rust to invalidate the binaries in the cache when changing contract versions.

maurolacy: Or the other way around, as you said: that the change in contract version alone didn't change the contract binary.

webmaster128:
0.10.0 is released on crates.io now. Can you please update this? Thank you

maurolacy:
Will do.

ethanfrey:
Looks good. Thanks!

PR #90, opened 2020-08-24
Fix the status on ending polls

levackt:
Bug slipped in with a refactor, the poll status is rejected even if passed.

### DIFF
--- a/voting/Cargo.toml
+++ b/voting/Cargo.toml
@@ -1,6 +1,6 @@
 [package]
 name = "cw-voting"
-version = "0.2.0"
+version = "0.2.1"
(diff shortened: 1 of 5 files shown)


voting/Cargo.toml
```diff
@@ -1,6 +1,6 @@
 [package]
 name = "cw-voting"
-version = "0.2.0"
+version = "0.2.1"
```
levackt, line 3: I wasn't sure if the version is still 0.2.0, feel free to edit or let me know.

ethanfrey (approved):
Good cacth,

Please request my review or mention me in the PR so I respond faster in the future.

PR #92, opened 2020-09-18
Move simple-option repo

orkunkl:
Related to https://github.com/CosmWasm/cosmwasm-plus/


.circleci/config.yml
```diff
@@ -270,3 +270,54 @@ jobs:
...
+          name: Add wasm32 target
+          command: rustup target add wasm32-unknown-unknown
+      - run:
+          name: Build
+          command: cargo wasm --locked
+      - run:
+          name: Unit tests
+          env: RUST_BACKTRACE=1
+          command: cargo unit-test --locked
+      - run:
```
ethanfrey, line 296: I believe there are not integration tests here, we can probably remove this line

ethanfrey: In fact this is not in cargo, meaning this should fail, meaning these tests are not yet run.

Please add simple-option to the jobs on top, let's see the test failure with this line, then remove that in another commit.

ethanfrey (approved):
Looks good.

But please fix up the CI setup and ensure all tests pass before mering

.circleci/config.yml
orkunkl: I moved it to the top but Circle CIwroker did not see the change.

orkunkl: I found what you mean :+1:

ethanfrey:
Nice

PR #95, opened 2020-09-23
simple-option: Migrate to cosmwasm v0.11.0

orkunkl:

@webmaster128 New error API is very nice thanks!

### DIFF
--- a/simple-option/Cargo.toml
+++ b/simple-option/Cargo.toml
@@ -31,11 +31,12 @@ overflow-checks = true
 backtraces = ["cosmwasm-std/backtraces"]
 
 [dependencies]
-cosmwasm-std = { version = "0.10.1" }
-cosmwasm-storage = { version = "0.10.1" }
+cosmwasm-std = { version = "0.11.0-alpha1" }
+cosmwasm-storage = { version = "0.11.0-alpha1" }
 schemars = "0.7"
 serde = { version = "1.0.103", default-features = false, features = ["derive"] }
 snafu = { version = "0.6.3" }
+thiserror = { version = "1.0.20" }
 
 [dev-dependencies]
-cosmwasm-schema = { version = "0.10.1" }
+cosmwasm-schema = { version = "0.11.0-alpha1" }
--- a/simple-option/Cargo.toml
+++ b/simple-option/Cargo.toml
@@ -31,11 +31,11 @@ overflow-checks = true
 backtraces = ["cosmwasm-std/backtraces"]
 
 [dependencies]
-cosmwasm-std = { version = "0.11.0-alpha2" }
-cosmwasm-storage = { version = "0.11.0-alpha2" }
+cosmwasm-std = { version = "0.11.0" }
+cosmwasm-storage = { version = "0.11.0" }
 schemars = "0.7"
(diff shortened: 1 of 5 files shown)


simple-option/src/contract.rs
```diff
@@ -219,11 +226,12 @@ mod tests {
     fn execute() {
         let mut deps = mock_dependencies(20, &[]);
 
-        let counter_offer = coins(40, "ETH");
+        let msg_counter_offer = coins(40, "ETH");
```
webmaster128, line 229: This is just an amount, so I think adding "msg" is misleading here

simple-option/src/error.rs
```diff
@@ -0,0 +1,30 @@
...
+    Unauthorized {},
+
+    #[error("must send exact counter offer (offer {offer:?}, counter_offer: {counter_offer:?})")]
+    CounterOfferMismatch {
+        offer: Vec<Coin>,
+        counter_offer: Vec<Coin>,
+    },
+
+    #[error("do not send funds with burn")]
+    FundSentWithBurn {},
```
webmaster128, line 29:
```suggestion
    FundsSentWithBurn {},
```

simple-option/Cargo.toml
```diff
@@ -31,11 +31,12 @@ overflow-checks = true
 backtraces = ["cosmwasm-std/backtraces"]
 
 [dependencies]
-cosmwasm-std = { version = "0.10.1" }
-cosmwasm-storage = { version = "0.10.1" }
+cosmwasm-std = { version = "0.11.0-alpha1" }
+cosmwasm-storage = { version = "0.11.0-alpha1" }
 schemars = "0.7"
 serde = { version = "1.0.103", default-features = false, features = ["derive"] }
 snafu = { version = "0.6.3" }
```
webmaster128, line 38: I think this dependency should be removed. We use thiserror instead of snafu, which is much simpler

simple-option/src/error.rs
```diff
@@ -0,0 +1,30 @@
...
+use thiserror::Error;
+
+#[derive(Error, Debug)]
+pub enum ContractError {
+    #[error("{}", original)]
+    Std {
+        // let thiserror implement From<StdError> for you
+        #[from]
+        original: StdError,
+    },
```
webmaster128, line 11: Question: do you think this way with the explocit `original` is easier to explain than mixed tuple/struct enum cases like:

```rust
pub enum ContractError {
    #[error("{}", original)]
    Std (#[from] StdError), // let thiserror implement From<StdError> for you
 
    #[error("expired option (expired {expired:?})")]
    OptionExpired { expired: u64 },

    // …
}
```

I am not sure which way to recommend as best practice. For an experiences Rust developer this is all the same.

orkunkl: To be honest, new learners will not ask `why this line is here?`. They will accept it as a requirement and copy-paste. If they want to learn why it's there `// let thiserror implement From<StdError> for you` line is giving enough tips for where to look. 
So tuple/struct is better, less code.

webmaster128: okay, this is what we documented here as well: https://github.com/CosmWasm/cosmwasm/blob/master/MIGRATING.md#011---010

PR #96, opened 2020-09-23
voting: Migrate to cosmwasm v0.11.0

orkunkl:


orkunkl:
Alright, integration tests are not working.

```
error[E0308]: mismatched types
   --> tests/integration.rs:873:9
    |
870 |     match res {
    |           --- this expression has type `cosmwasm_std::results::contract_result::ContractResult<cosmwasm_std::results::handle::HandleResponse>`
...
873 |         Err(e) => panic!("Unexpected error: {:?}", e),
    |         ^^^^^^ expected enum `cosmwasm_std::results::contract_result::ContractResult`, found enum `std::result::Result`
    |
    = note: expected enum `cosmwasm_std::results::contract_result::ContractResult<cosmwasm_std::results::handle::HandleResponse>`
               found enum `std::result::Result<_, _>`

error: aborting due to 36 previous errors; 1 warning emitted
```
@webmaster128 any ideas?

webmaster128:
When matching the ContractResult, you need to prefix the result:

```rust
match result {
    ContractResult::Ok(value) => {}
    ContractResult::Err(err) => {}
}
```

Before this was a standard `Result` where you can magically skip the `Result::`

orkunkl:
So far ContractError design worked well for unit tests but got weird in integration tests.
Getting this error:
```
error[E0308]: mismatched types
   --> tests/integration.rs:128:29
    |
126 |     match res {
    |           --- this expression has type `cosmwasm_std::results::contract_result::ContractResult<cosmwasm_std::results::handle::HandleResponse<_>>`
127 |         ContractResult::Ok(_) => panic!("Must return error"),
128 |         ContractResult::Err(ContractError::PollQuorumPercentageMismatch { .. }) => {}
    |                             ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ expected struct `std::string::String`, found enum `cw_voting::error::ContractError`
```
I tried to make it this expecting string will be converted to ContractError magically.

```rust
    match res {
        ContractResult::Err(e) => {
            match e {
                ContractError::PollNotExist{} => {}
                _ => panic!("Unexpected error: {:?}", e)
            }
        },
        _ => panic!("Must return error"),
    }
```
Of course, I got error telling `ContractError` is not a string.
Then I tried to parse error string to error assuming `thiserror` would derive `into` and `from` for the contract error.
```rust
    match res {
        ContractResult::Err(e) => {
            let err: ContractError = e.into();
            match err {
                ContractError::PollNotExist{} => {}
                _ => panic!("Unexpected error: {:?}", e)
            }
        },
        _ => panic!("Must return error"),
    }
```

```
error[E0277]: the trait bound `cw_voting::error::ContractError: std::convert::From<std::string::String>` is not satisfied
   --> tests/integration.rs:107:40
    |
107 |             let err: ContractError = e.into();
    |                                        ^^^^ the trait `std::convert::From<std::string::String>` is not implemented for `cw_voting::error::ContractError`
    |
    = help: the following implementations were found:
              <cw_voting::error::ContractError as std::convert::From<cosmwasm_std::errors::std_error::StdError>>
    = note: required because of the requirements on the impl of `std::convert::Into<cw_voting::error::ContractError>` for `std::string::String`
```
No success 😢 

Suggestions?

webmaster128:
> Suggestions?

Yeah, in integration tests you don't get structured errors. As soon as they leave the contract into the VM, they are converted to strings. So in integration tests you have to check strings. But here you don't need to write a test for every single edge case. Those are easier to maintain as unit tests.

voting/schema/poll_response.json
```diff
@@ -47,6 +47,7 @@
       "type": "string"
     },
     "PollStatus": {
+      "type": "string",
```
orkunkl, line 50: Is this change expected?

voting/src/error.rs
```diff
@@ -0,0 +1,53 @@
+use cosmwasm_std::{CanonicalAddr, StdError};
+use thiserror::Error;
+
+#[derive(Error, Debug)]
+pub enum ContractError {
+    #[error("{0}")]
+    Std(#[from] StdError),
+
+    #[error("insufficient funds sent")]
+    InsufficientFunds {},
```
webmaster128, line 10:
```suggestion
    InsufficientFundsSent {},
```

Otherwise it sounds like the balance is not enought

voting/src/error.rs
```diff
@@ -0,0 +1,53 @@
...
+
+    #[error("excessive withdrawal amount (max_amount {max_amount:?})")]
```
webmaster128, line 12: Why are you adding those question marks in the template? They probably give you the Debug format instead of the DIsplay format.

voting/src/error.rs
```diff
@@ -0,0 +1,53 @@
...
+    ExcessiveWithdraw { max_amount: u128 },
+
+    #[error("description too short (minimum description length {min_desc_length:?})")]
+    DescriptionTooShort { min_desc_length: usize },
```
webmaster128, line 16: As a general best practice, I'd avoid using `usize` in all contract code. This is 64 bit on a developer machine and in unit tests but 32 bits in Wasm. Better use `u64` and cast `usize` to `u64` when necessary.

voting/src/lib.rs
```diff
@@ -1,5 +1,6 @@
 pub mod coin_helpers;
 pub mod contract;
+pub mod error;
```
webmaster128, line 3: I don't think the `error` module should be public, since the types are only used internally.

voting/src/error.rs
orkunkl: From [thiserror docs](https://docs.rs/thiserror/1.0.20/thiserror)
```
A Display impl is generated for your error if you provide #[error("...")] messages on the struct or each variant of your enum, as shown above in the example.

The messages support a shorthand for interpolating fields from the error.

#[error("{var}")] ⟶ write!("{}", self.var)
#[error("{0}")] ⟶ write!("{}", self.0)
#[error("{var:?}")] ⟶ write!("{:?}", self.var)
#[error("{0:?}")] ⟶ write!("{:?}", self.0)
```

webmaster128: Jupp, let's use the first version:

```
#[error("{var}")] ⟶ write!("{}", self.var)
```

voting/src/error.rs
orkunkl: I could not find a clear way to convert usize to u64. https://users.rust-lang.org/t/cant-convert-usize-to-u64/6243/6. It cannot be safely converted. usize is convertable only to u8.

voting/src/error.rs
orkunkl: Sounds good.

voting/src/error.rs
webmaster128: You can cast `let my_var = xyz.len() /* returns usize*/ as u64`, which is safe on all our platforms. What this guy is saying is you can't safely convert u64 to usize because usize can be smaller.

webmaster128: Here in the error usize could be used. But people quickly copy/paste code around and as soon as usize makes it into a request or response message, things fall apart because usize cannot be serialized/deserialized.

PR #99, opened 2020-10-05
escrow: Migrate to cosmwasm v0.11.0

orkunkl:

### DIFF
--- a/escrow/src/contract.rs
+++ b/escrow/src/contract.rs
@@ -23,8 +23,8 @@ pub fn init<S: Storage, A: Api, Q: Querier>(
 
     if state.is_expired(&env) {
         return Err(ContractError::Expired {
-            end_height: env.block.height,
-            end_time: env.block.time,
+            end_height: msg.end_height,
+            end_time: msg.end_time,
         });
     }
 
@@ -59,8 +59,8 @@ fn try_approve<S: Storage, A: Api, Q: Querier>(
     // throws error if state is expired
     if state.is_expired(&env) {
         return Err(ContractError::Expired {
-            end_height: env.block.height,
-            end_time: env.block.time,
+            end_height: state.end_height,
+            end_time: state.end_time,
         });
     }
 
--- a/escrow/src/contract.rs
+++ b/escrow/src/contract.rs
@@ -1,8 +1,9 @@
 use cosmwasm_std::{
-    log, to_binary, Api, BankMsg, Binary, CanonicalAddr, Coin, CosmosMsg, Env, Extern,
-    HandleResponse, HandleResult, InitResponse, InitResult, Querier, StdError, StdResult, Storage,
+    attr, to_binary, Api, BankMsg, Binary, CanonicalAddr, Coin, CosmosMsg, Env, Extern,
+    HandleResponse, InitResponse, Querier, StdResult, Storage,
 };
 
+use crate::error::ContractError;
 use crate::msg::{ArbiterResponse, HandleMsg, InitMsg, QueryMsg};
 use crate::state::{config, config_read, State};
 
@@ -10,7 +11,7 @@ pub fn init<S: Storage, A: Api, Q: Querier>(
     deps: &mut Extern<S, A, Q>,
     env: Env,
     msg: InitMsg,
-) -> InitResult {
+) -> Result<InitResponse, ContractError> {
     let state = State {
         arbiter: deps.api.canonical_address(&msg.arbiter)?,
         recipient: deps.api.canonical_address(&msg.recipient)?,
@@ -18,19 +19,20 @@ pub fn init<S: Storage, A: Api, Q: Querier>(
         end_height: msg.end_height,
         end_time: msg.end_time,
     };
-    if state.is_expired(&env) {
-        Err(StdError::generic_err("creating expired escrow"))
-    } else {
-        config(&mut deps.storage).save(&state)?;
-        Ok(InitResponse::default())
+    match state.is_expired(&env) {
+        ContractError::EscrowNotExpired {} => {}
+        err => return Err(err),
     }
+
+    config(&mut deps.storage).save(&state)?;
+    Ok(InitResponse::default())
 }
 
 pub fn handle<S: Storage, A: Api, Q: Querier>(
     deps: &mut Extern<S, A, Q>,
     env: Env,
     msg: HandleMsg,
-) -> HandleResult {
+) -> Result<HandleResponse, ContractError> {
     let state = config_read(&deps.storage).load()?;
     match msg {
         HandleMsg::Approve { quantity } => try_approve(deps, env, state, quantity),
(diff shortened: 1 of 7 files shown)


escrow/src/error.rs
```diff
@@ -0,0 +1,20 @@
+use cosmwasm_std::StdError;
+use thiserror::Error;
+
+#[derive(Error, Debug)]
+pub enum ContractError {
+    #[error("{0}")]
+    Std(#[from] StdError),
+
+    #[error("unauthorized")]
```
maurolacy, line 9: This is a matter of taste, but I prefer capitalization of error messages, i. e. `"Unauthorized"`.

escrow/src/error.rs
```diff
@@ -0,0 +1,20 @@
...
+    Unauthorized {},
+
+    #[error("escrow expired (end_height {end_height})")]
+    EscrowExpiredHeight { end_height: u64 },
```
maurolacy, line 13: Again a matter of taste, but I think `Escrow` in the error enum is redundant. I would just use `ExpiredHeight`, and so on.

escrow/src/contract.rs
```diff
@@ -1,122 +1,132 @@
...
+        source: deps.api.canonical_address(&info.sender)?,
         end_height: msg.end_height,
         end_time: msg.end_time,
     };
-    if state.is_expired(&env) {
-        Err(StdError::generic_err("creating expired escrow"))
-    } else {
-        config(&mut deps.storage).save(&state)?;
-        Ok(InitResponse::default())
+    match state.is_expired(&env) {
```
maurolacy, line 23: Another matter of taste, but I like the `if` better. I think the intention is clear.

Also, I try to avoid doing structural changes when migrating; unless strictly necessary, fixing something, etc.

escrow/src/contract.rs
```diff
@@ -1,122 +1,132 @@
...
 
-        send_tokens(
-            &deps.api,
-            &deps.api.canonical_address(&env.contract.address)?,
-            &state.recipient,
-            amount,
-            "approve",
-        )
+    // throws error if state is expired
+    match state.is_expired(&env) {
```
maurolacy, line 57: It's you call, but, same here.

These changes also hinder reviewing, as I need to pay attention / verify stuff that is not strictly related to the migration. You can also inadvertently introduce subtle bugs, etc.

Now, for me you can leave these changes (if you're sure they are correct). I'm just mentioning these things as general guidelines.

escrow/src/contract.rs
```diff
@@ -1,122 +1,132 @@
...
-        // https://github.com/CosmWasm/wasmd/blob/master/x/wasm/internal/keeper/keeper.go#L185-L192
-        let balance = deps.querier.query_all_balances(&env.contract.address)?;
-        send_tokens(
-            &deps.api,
-            &deps.api.canonical_address(&env.contract.address)?,
-            &state.source,
-            balance,
-            "refund",
-        )
+    let err = state.is_expired(&env);
```
maurolacy, line 87: I can't help but mention again that the original version looks clearer to me.

escrow/src/contract.rs
```diff
@@ -1,122 +1,132 @@
...
 }
 
 // this is a helper to move the tokens, so the business logic is easy to read
-fn send_tokens<A: Api>(
-    api: &A,
-    from_address: &CanonicalAddr,
-    to_address: &CanonicalAddr,
+fn send_tokens(
+    from_address: HumanAddr,
+    to_address: HumanAddr,
```
maurolacy, line 109: Nice.

escrow/src/contract.rs
```diff
@@ -147,20 +157,33 @@ mod tests {
...
+                height,
+                time,
+                ..Default::default()
+            },
+            contract: ContractInfo {
+                address: HumanAddr::from(MOCK_CONTRACT_ADDR),
+            },
+        };
+        let info = mock_info(signer, sent);
+        return (env, info);
```
maurolacy, line 177: As there are no dependencies between `mock_env_height()` and `mock_info()`, there's no real need to handle them together.

escrow/src/contract.rs
```diff
@@ -188,20 +211,20 @@ mod tests {
...
+        let mut deps = mock_dependencies(&[]);
 
         let msg = init_msg_expire_by_height(1000);
-        let env = mock_env_height("creator", &coins(1000, "earth"), 1001, 0);
-        let res = init(&mut deps, env, msg);
+        let (env, info) = mock_env_info_height("creator", &coins(1000, "earth"), 1001, 0);
+        let res = init(&mut deps, env, info, msg);
         match res.unwrap_err() {
-            StdError::GenericErr { msg, .. } => assert_eq!(msg, "creating expired escrow"),
+            ContractError::EscrowExpiredHeight { .. } => {}
```
maurolacy, line 220: Nice! :-)

escrow/src/state.rs
```diff
@@ -15,20 +16,23 @@ pub struct State {
 }
 
 impl State {
-    pub fn is_expired(&self, env: &Env) -> bool {
+    pub fn is_expired(&self, env: &Env) -> ContractError {
```
maurolacy, line 19: I don't think this is a good idea. You are changing a boolean check by a function returning an error. And using the error enums to signal a condition that's not (not necessarily) an error.

escrow/tests/integration.rs
```diff
@@ -86,155 +94,3 @@ fn proper_initialization() {
...
-    let msg = handle_res.messages.get(0).expect("no message");
-    assert_eq!(
-        msg,
-        &CosmosMsg::Bank(BankMsg::Send {
-            from_address: HumanAddr::from("cosmos2contract"),
-            to_address: HumanAddr::from("creator"),
-            amount: coins(1000, "earth"),
-        })
-    );
-}
```
maurolacy, line 240: I think this is OK (removing integration tests, as discussed) but would have been better to do it as part of another PR. Or, at least, in a different (atomic) commit, so the intention / objective is clearer.

maurolacy (commented):
lgtm overall.

escrow/src/contract.rs
webmaster128: I agree. When a function starts with `is_`, this indicates a boolean return type.

maurolacy (approved):
LGTM.

escrow/src/error.rs
```diff
@@ -0,0 +1,17 @@
...
+#[derive(Error, Debug)]
+pub enum ContractError {
+    #[error("{0}")]
+    Std(#[from] StdError),
+
+    #[error("unauthorized")]
+    Unauthorized {},
+
+    #[error("escrow expired (end_height {end_height})")]
+    Expired { end_height: u64, end_time: u64 },
```
webmaster128, line 13: `end_height` and `end_time` of the escrow are both optional, as far as I know.

We can either wrap this on `Option` or add documention to the fields saying that e.g. `0` is used when unset.

escrow/src/contract.rs
```diff
@@ -1,122 +1,133 @@
...
     };
+
     if state.is_expired(&env) {
-        Err(StdError::generic_err("creating expired escrow"))
-    } else {
-        config(&mut deps.storage).save(&state)?;
-        Ok(InitResponse::default())
+        return Err(ContractError::Expired {
+            end_height: env.block.height,
+            end_time: env.block.time,
```
webmaster128, line 27: This sets the current height and current time, not the end of the escrow. The values you look for are in `state`

escrow/src/contract.rs
```diff
@@ -1,122 +1,133 @@
...
-            &deps.api.canonical_address(&env.contract.address)?,
-            &state.recipient,
-            amount,
-            "approve",
-        )
+    // throws error if state is expired
+    if state.is_expired(&env) {
+        return Err(ContractError::Expired {
+            end_height: env.block.height,
+            end_time: env.block.time,
```
webmaster128, line 63: As above, those are not the values of the escrow

escrow/src/contract.rs
maurolacy: You're right!

escrow/src/contract.rs
```diff
@@ -23,8 +23,8 @@ pub fn init<S: Storage, A: Api, Q: Querier>(
 
     if state.is_expired(&env) {
         return Err(ContractError::Expired {
-            end_height: env.block.height,
-            end_time: env.block.time,
+            end_height: msg.end_height,
+            end_time: msg.end_time,
```
maurolacy, line 27: You can use `state` here too for clarity, but it's the same.

escrow/src/error.rs
```diff
@@ -9,8 +9,11 @@ pub enum ContractError {
     #[error("unauthorized")]
     Unauthorized {},
 
-    #[error("escrow expired (end_height {end_height})")]
-    Expired { end_height: u64, end_time: u64 },
+    #[error("escrow expired (end_height {end_height:?} end_time {end_time:?})")]
```
maurolacy: I guess these will show up as `None` when unset. Which is a good alternative IMO.

maurolacy (approved):
This is correct now, AFAIK.

escrow/src/error.rs
webmaster128: Yes. For `Option` you need to use the `Debug` printing (thus the `:?`) because `Display` is not implemented

webmaster128:
Thanks!

PR #111, opened 2020-11-14
CI build release contracts

### LINKED ISSUE #87: Auto-build contracts on releases
- Use rust-optimizer 0.10.0
- Push them to github on each tagged release.
- Use https://github.com/CosmWasm/cosmwasm-plus/pull/29 as reference

maurolacy:
Closes #87 .


.circleci/config.yml
```diff
@@ -267,3 +275,50 @@ jobs:
...
+            # copy code into this volume
+            docker cp . with_code:/code
+      - run:
+          name: Build development contracts
+          command: |
+            docker run --volumes-from with_code cosmwasm/rust-optimizer:0.10.5 ./erc20
+            docker run --volumes-from with_code cosmwasm/rust-optimizer:0.10.5 ./escrow
+            docker run --volumes-from with_code cosmwasm/rust-optimizer:0.10.5 ./nameservice
+            docker run --volumes-from with_code cosmwasm/rust-optimizer:0.10.5 ./simple-option
+            docker run --volumes-from with_code cosmwasm/rust-optimizer:0.10.5 ./voting
```
maurolacy, line 305: This can be improved upon by allowing rust-optimizer to accept a list of directories. I'll create the issue, and implement the (probably small) changes.

.circleci/config.yml
```diff
@@ -267,3 +275,46 @@ jobs:
...
+          name: Show data
+          command: |
+            ls -l artifacts
+            cat artifacts/checksums.txt
+      - run:
+          name: Publish artifacts on GitHub
+          command: |
+            TAG="$CIRCLE_TAG"
+            TITLE="$TAG"
+            BODY="Attached there are some build artifacts generated at this tag. Those are for development purposes only! Please use crates.io to find the packages of this release."
```
webmaster128, line 314: Can we use a text that better matches this repo? Better write less or nothing than the text from cosmwasm.

It is the description shown here. :
Bildschirmfoto 2020-11-16 um 10 17 59

.circleci/config.yml
```diff
@@ -267,3 +275,46 @@ jobs:
...
+          name: Prepare volume with source code
+          command: |
+            # create a dummy container which will hold a volume with config
+            docker create -v /code --name with_code alpine /bin/true
+            # copy code into this volume
+            docker cp . with_code:/code
+      - run:
+          name: Build development contracts
+          command: |
+            docker run --volumes-from with_code cosmwasm/rust-optimizer:0.10.6 ./*/
```
webmaster128, line 301: What do we gain from using a single call here? Primarily simplifies caller scripts but does not speed up compilation, right?

.circleci/config.yml
```diff
@@ -267,3 +275,46 @@ jobs:
             - /usr/local/cargo/registry
             - target
           key: cargocache-voting-rust:1.47.0-{{ checksum "Cargo.lock" }}
+
+  # This job is based in cosmwasm-plus/.circleci/config.yml
+  build_and_upload_contracts:
+    docker:
+      # Image from https://github.com/cibuilds/github, based on alpine
+      - image: cibuilds/github:0.13
```
webmaster128, line 283: FYI: In case we either dron't trust this image or want a dofferent one, we can easily install this tool in a different container as long as the container can execute `go get`: https://github.com/CosmWasm/wasmvm/blob/v0.12.0-alpha3/.circleci/config.yml#L120-L128. However, I think this is good for now.

simple-option/Cargo.toml
```diff
@@ -1,6 +1,6 @@
 [package]
 name = "simple-option"
-version = "0.1.0"
+version = "0.2.0"
```
webmaster128, line 3: Since we build all contracts on the tags now, we better use consistent versioning for the whole repo.

Can you adapt the tool https://github.com/CosmWasm/cosmwasm/blob/master/devtools/set_version.sh to set one version in all crates? In this repo I renamed to tools folder to `.devtools` in order to not incluse it when globbing `./*` or `./*/`.

webmaster128 (commented):
Looks good, thanks. Can you also:

- Remove all `*.wasm`, `hash.txt` and `checksums.txt` files from the repo
- Remove `exclude =` statements in Cargo.toml
- Update the `### Note on build results` text block in `Publishing.md`s

.circleci/config.yml
maurolacy: Yes, it's faster. It's also more general, as we don't need to explicitly list the different directories / contracts.

webmaster128 (approved):
Beautiful. Let me try tagging this :)

webmaster128:
Nice one: https://github.com/CosmWasm/cosmwasm-examples/releases/tag/v0.8.0

PR #116, opened 2021-01-21
Upgrade to cw 0.13.2

orkunkl:
Note: integration tests crashes, I tried to resolve the issues but could not. I can solve them if you help or just disable on CI level.


ethanfrey:
I think I fixed three of these tests. You can try to use the same pattern for the others (or I will later)

PR #120, opened 2021-05-06
Upgrade escrow to 0.14

orkunkl:
Related to https://github.com/CosmWasm/cosmwasm-examples/issues/121
This is required for updating docs to 0.14


contracts/escrow/src/msg.rs
```diff
@@ -1,11 +1,11 @@
...
 use schemars::JsonSchema;
 use serde::{Deserialize, Serialize};
 
 #[derive(Serialize, Deserialize, Clone, Debug, PartialEq, JsonSchema)]
-pub struct InitMsg {
-    pub arbiter: HumanAddr,
-    pub recipient: HumanAddr,
+pub struct InstantiateMsg {
+    pub arbiter: Addr,
+    pub recipient: Addr,
```
webmaster128, line 8: The message types must not use `Addr` because this is data from users that is not yet validated. We should use `String` here

contracts/escrow/src/state.rs
```diff
@@ -23,7 +23,7 @@ impl State {
         }
 
         if let Some(end_time) = self.end_time {
-            if env.block.time > end_time {
+            if env.block.time.nanos() > end_time {
```
webmaster128, line 26: `State.end_time` is in seconds but the block time changes to nanoseconds. We need something like

```suggestion
            if env.block.time.nanos() > end_time * 1000 {
```

contracts/escrow/src/contract.rs
```diff
@@ -1,22 +1,22 @@
...
-) -> Result<InitResponse, ContractError> {
+    msg: InstantiateMsg,
+) -> Result<Response, ContractError> {
     let state = State {
-        arbiter: deps.api.canonical_address(&msg.arbiter)?,
-        recipient: deps.api.canonical_address(&msg.recipient)?,
-        source: deps.api.canonical_address(&info.sender)?,
+        arbiter: deps.api.addr_canonicalize(&msg.arbiter.as_str())?,
+        recipient: deps.api.addr_canonicalize(&msg.recipient.as_str())?,
+        source: deps.api.addr_canonicalize(&info.sender.as_str())?,
```
webmaster128, line 19: We have a new `deps.api.addr_validate` that allows us to use `Addr` in `State` and `String` in the message. Then no transformation is required. We just validate the user input.

PR #122, opened 2021-05-12
Upgrade simple-option to 0.14

orkunkl:
Related to https://github.com/CosmWasm/cosmwasm-examples/issues/121


maurolacy (approved):
lgtm. There are a couple of reference to `handle` and `init` in the docs (`Developing.md` and `Importing.md`) that you would want to change too.

webmaster128:
> There are a couple of reference to `handle` and `init` in the docs (`Developing.md` and `Importing.md`) that you would want to change too.

I think it's better to delete them. Maintaining them in each contract folder is an impossible task.

orkunkl:
@webmaster128 those markdown files, come from templates. I will fix it there but you are right these are unnecessary, will delete.

contracts/simple-option/src/contract.rs
```diff
@@ -22,56 +22,57 @@ pub fn init(
...
-    state.owner = recipient.clone();
+    state.owner = deps.api.addr_validate(&recipient)?;
     config(deps.storage).save(&state)?;
 
-    let mut res = Context::new();
+    let mut res = Response::new();
     res.add_attribute("action", "transfer");
     res.add_attribute("owner", recipient);
-    Ok(res.into())
+    Ok(res)
```
webmaster128, line 68: So much simpler now :)

Issue #121, opened 2021-05-07
Upgrade to 0.14


orkunkl:
- [x] escrow
- [x] erc20
- [x] name-service
- [x] simple-option
- [x] voting

PR #124, opened 2021-05-12
Upgrade erc20 to 0.14

orkunkl:
Related to #121


contracts/erc20/src/contract.rs
```diff
@@ -319,6 +326,7 @@ fn read_allowance(
     read_u128(&owner_store, spender.as_slice())
 }
 
+#[allow(clippy::unnecessary_wraps)]
```
maurolacy, line 329: You can avoid this by removing `write_allowance`'s return value, and modifying calls to it. (remove the subsequent question mark).

Also, probably out of scope here, but if you run clippy with `--all-targets`, there are a bunch of `clone()`s that can be removed.

contracts/erc20/src/contract.rs
```diff
@@ -384,50 +392,45 @@ mod tests {
         return bytes_to_u128(&data).unwrap();
     }
 
-    fn get_balance(api: &dyn Api, storage: &dyn Storage, address: &HumanAddr) -> u128 {
+    fn get_balance(api: &dyn Api, storage: &dyn Storage, address: &Addr) -> u128 {
         let address_key = api
-            .canonical_address(address)
+            .addr_canonicalize(address.as_str())
```
maurolacy, line 397: You can avoid this conversion entirely, and just use `address` as directly key.

contracts/erc20/src/contract.rs
```diff
@@ -384,50 +392,45 @@ mod tests {
...
-    fn get_allowance(
-        api: &dyn Api,
-        storage: &dyn Storage,
-        owner: &HumanAddr,
-        spender: &HumanAddr,
-    ) -> u128 {
+    fn get_allowance(api: &dyn Api, storage: &dyn Storage, owner: &Addr, spender: &Addr) -> u128 {
         let owner_raw_address = api
-            .canonical_address(owner)
+            .addr_canonicalize(owner.as_str())
```
maurolacy, line 404: Same here.

contracts/erc20/src/contract.rs
```diff
@@ -384,50 +392,45 @@ mod tests {
...
             .expect("canonical_address failed");
         let spender_raw_address = api
-            .canonical_address(spender)
+            .addr_canonicalize(spender.as_str())
```
maurolacy, line 407: And here. Just remove all references to / uses of "canonical addresses".

contracts/erc20/src/contract.rs
```diff
@@ -384,50 +392,45 @@ mod tests {
         return bytes_to_u128(&data).unwrap();
     }
 
-    fn get_balance(api: &dyn Api, storage: &dyn Storage, address: &HumanAddr) -> u128 {
+    fn get_balance(api: &dyn Api, storage: &dyn Storage, address: &Addr) -> u128 {
```
maurolacy, line 395: Suggestion: Given that these are helper functions for tests, you can save a lot in conversions by just using `&str` for addresses.

contracts/erc20/src/msg.rs
```diff
@@ -1,16 +1,16 @@
 use schemars::JsonSchema;
 use serde::{Deserialize, Serialize};
 
-use cosmwasm_std::{HumanAddr, Uint128};
+use cosmwasm_std::{Addr, Uint128};
 
 #[derive(Serialize, Deserialize, Clone, PartialEq, JsonSchema)]
 pub struct InitialBalance {
-    pub address: HumanAddr,
+    pub address: Addr,
```
maurolacy, line 8: This should be of type `String`, as this struct is part of a message that can contain unvalidated addresses. When changing this, a lot of conversions (mostly in tests) will simplify / disappear.

contracts/erc20/src/contract.rs
```diff
@@ -384,50 +392,45 @@ mod tests {
...
         fn works() {
             let mut deps = mock_dependencies(&[]);
-            let init_msg = InitMsg {
+            let instantiate_msg = InstantiateMsg {
                 name: "Cash Token".to_string(),
                 symbol: "CASH".to_string(),
                 decimals: 9,
                 initial_balances: [InitialBalance {
-                    address: HumanAddr("addr0000".to_string()),
+                    address: Addr::unchecked("addr0000".to_string()),
```
maurolacy, line 427: This and many others will simplify to `"addr0000".to_string()` when changing `InitialBalance::address`'s type to `String`.

contracts/erc20/src/contract.rs
```diff
@@ -357,14 +365,14 @@ mod tests {
     use super::*;
     use crate::msg::InitialBalance;
     use cosmwasm_std::testing::{mock_dependencies, mock_env, mock_info};
-    use cosmwasm_std::{from_slice, Api, Env, HumanAddr, MessageInfo, Storage, Uint128};
+    use cosmwasm_std::{from_slice, Addr, Api, Env, MessageInfo, Storage, Timestamp, Uint128};
     use cosmwasm_storage::ReadonlyPrefixedStorage;
 
-    fn mock_env_height(signer: &HumanAddr, height: u64, time: u64) -> (Env, MessageInfo) {
+    fn mock_env_height(signer: &Addr, height: u64, time: u64) -> (Env, MessageInfo) {
```
maurolacy, line 371: Better to use `&str` for `signer` I guess.

contracts/erc20/src/contract.rs
```diff
@@ -357,14 +365,14 @@ mod tests {
...
         let mut env = mock_env();
-        let info = mock_info(signer, &[]);
+        let info = mock_info(signer.as_str(), &[]);
         env.block.height = height;
-        env.block.time = time;
+        env.block.time = Timestamp::from_nanos(time);
```
maurolacy, line 375: Not really relevant for these tests, but I think `from_seconds` makes more sense here.

maurolacy (changes requested):
Please address the comments, and I'll re-review, and finish reviewing tests.

contracts/erc20/src/contract.rs
orkunkl: I think making helpers return result could be a good practice, in case api changes and something that could return error inserted in write_allowance, no need to introduce huge changes

contracts/erc20/src/contract.rs
orkunkl: yeah introducs lots of change now.

maurolacy:
There's still an `addr_canonicalize()` in `src/contract.rs`, in the `token_balance()` fn that can be easily removed.

orkunkl:
I will disable integration tests because vm BackendApi is not updated

contracts/erc20/src/contract.rs
```diff
@@ -296,38 +283,39 @@ pub fn bytes_to_u128(data: &[u8]) -> Result<u128, ContractError> {
 
 // Reads 16 byte storage value into u128
 // Returns zero if key does not exist. Errors if data found that is not 16 bytes
-pub fn read_u128(store: &ReadonlyPrefixedStorage, key: &[u8]) -> Result<u128, ContractError> {
-    let result = store.get(key);
+pub fn read_u128(store: &ReadonlyPrefixedStorage, key: &Addr) -> Result<u128, ContractError> {
```
maurolacy, line 286: Just a detail: I would keep this function signature as it is, and convert to slice outside.

contracts/erc20/src/contract.rs
```diff
@@ -1388,35 +1365,35 @@ mod tests {
...
 
-        fn address(index: u8) -> HumanAddr {
+        fn address(index: u8) -> Addr {
             match index {
-                0 => HumanAddr("addr0000".to_string()), // contract initializer
-                1 => HumanAddr("addr1111".to_string()),
-                2 => HumanAddr("addr4321".to_string()),
-                3 => HumanAddr("addr5432".to_string()),
-                4 => HumanAddr("addr6543".to_string()),
+                0 => Addr::unchecked("addr0000".to_string()), // contract instantiateializer
```
maurolacy, line 1372: :-)
```suggestion
                0 => Addr::unchecked("addr0000".to_string()), // contract instantiation

File: contracts/erc20/src/contract.rs [hunk 12]
```diff
@@ -1440,12 +1417,12 @@ mod tests {
...
-            let init_msg = make_init_msg();
-            let (env, info) = mock_env_height(&address(0), 450, 550);
-            let res = init(deps.as_mut(), env.clone(), info, init_msg).unwrap();
+            let instantiate_msg = make_instantiate_msg();
+            let (env, info) = mock_env_height(&address(0).as_str(), 450, 550);
+            let res = instantiate(deps.as_mut(), env.clone(), info, instantiate_msg).unwrap();
             assert_eq!(0, res.messages.len());
             let query_msg = QueryMsg::Balance {
-                address: address(4), // only indices 1, 2, 3 are initialized
+                address: address(4).to_string(), // only indices 1, 2, 3 are instantiateialized
```
maurolacy, line 1425:
```suggestion
                address: address(4).to_string(), // only indices 1, 2, 3 are initialized
```

maurolacy (approved):
lgtm.

maurolacy:
I've seen integration tests are failing (when ran manually). Let's either fix them, or delete them.

PR #128, opened 2021-05-27
Use .as_str().as_bytes() consistently to get bytes of Addr

webmaster128:
Use the explicit methods names to convert `Addr`/`&Addr` -> `&str` -> `&[u8]`

### DIFF
--- a/contracts/erc20/src/contract.rs
+++ b/contracts/erc20/src/contract.rs
@@ -260,7 +260,7 @@ fn perform_transfer(
         });
     }
     from_balance -= amount;
-    balances_store.set(from.as_str().as_ref(), &from_balance.to_be_bytes());
+    balances_store.set(from.as_str().as_bytes(), &from_balance.to_be_bytes());
 
     let mut to_balance = match balances_store.get(to.as_str().as_bytes()) {
         Some(data) => bytes_to_u128(&data),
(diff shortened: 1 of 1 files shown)


contracts/erc20/src/contract.rs
```diff
@@ -260,7 +260,7 @@ fn perform_transfer(
         });
     }
     from_balance -= amount;
-    balances_store.set(from.as_str().as_ref(), &from_balance.to_be_bytes());
+    balances_store.set(from.as_str().as_bytes(), &from_balance.to_be_bytes());
```
orkunkl, line 263: I've used `.as_str().as_bytes()` code over and over. Is it possible to implement `Into<&[u8]>` or `.as_key` for Addr?

orkunkl (approved):
LGTM

contracts/erc20/src/contract.rs
webmaster128: Implementing `Into` does not make much sense as you would lose an owned instance just to get a read-only reference to the data.

However, I understand the desire. Ethan wanted this as well already. We can add `Addr::as_bytes()` directly, which is implememented via `Addr`/`&Addr` -> `&str` -> `&[u8]`. I was a bit reserved doing that as I was not sure if it is intuitive what such an API does. There are many ways to represent an address as binary. But yeah, if you want `Addr::as_bytes()` we can add it.

webmaster128: https://github.com/CosmWasm/cosmwasm/issues/923

orkunkl: Shortening a repetitive command improves Dev UX. Good to add it.

PR #131, opened 2021-08-12
Update max decimal limit in example ERC20 contract

rellfy:
The max doesn't have to be 18 decimals, unless there's some restriction I'm unaware of?
Nice project by the way.

### DIFF
--- a/contracts/erc20/src/contract.rs
+++ b/contracts/erc20/src/contract.rs
@@ -41,7 +41,7 @@ pub fn instantiate(
     if !is_valid_symbol(&msg.symbol) {
         return Err(ContractError::TickerWrongSymbolFormat {});
     }
-    if msg.decimals > u8::MAX {
+    if msg.decimals > u8::MAX - 1 {
         return Err(ContractError::DecimalsExceeded {});
     }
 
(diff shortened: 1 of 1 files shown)


webmaster128:
You are right, it doesn't have to be that value. But there are reasons for using it.

18 is a very typical value used in blockchain systems. For ERC20 contracts this is the default. I've seen many examples with fewer decimals but none with more.

The decimal amounts are stured as unsigned 128 bit integers, including all decimal digits. The largest uint128 value is

```
2**128-1
340282366920938463463374607431768211455
```

which translates to the decimal `340282366920938463463.374607431768211455`. So you can represent a supply of up to 340 quintillion with that. When we increase the decimals, the dot moves to the left.

```
18: 340282366920938463463.374607431768211455 (340 quintillion)
21: 340282366920938463.463374607431768211455 (340 quadrillion)
24: 340282366920938.463463374607431768211455 (340 trillion)
27: 340282366920.938463463374607431768211455 (340 billion)
30: 340282366.920938463463374607431768211455 (340 million)
33: 340282.366920938463463374607431768211455 (340 thousand)
36: 340.282366920938463463374607431768211455 (340)
```

A typical value for token supplies is tens to hundreds of million. If you want to be able to represent that, you cannot exceed 30 decimal places.

So yeah, 18 is more limiting than necessary. But 255 is not a practical limit. 30 or 40 is probably somewhat reasonable.

mikedotexe:
Looks like that example is removed and this PR can be closed

webmaster128:
Right. cw20 in cw-plus made this example obsolete.

PR #145, opened 2022-05-23
Remove erc20

orkunkl:
Removing erc20 sample. It is confusing for people coming from eth world. Their natural action is using erc20 instead of cw20. Best to delete this sample.


ethanfrey (approved):
lgtm

